PR #84, opened 2025-07-09
Implement cropper modal

### LINKED ISSUE #74: (SP:) Implement Cropper modal
Implement Cropper modal for changing pictured based on specific crop size (using the library that would suit best)

AC:
- matches design in figma
- crops an image correctly
- loads the needed picture
- saves the file when saving

Image

lizabre:
Implemented cropper modal for images
- added tests
- returns as link and blob 
- crops based on set pixels
- used react-image-crop

Example

https://github.com/user-attachments/assets/a30acd2b-688a-44fa-b857-21a8c749760b


app/shared/components/design-system/cropper-modal/CropperModal.tsx
```diff
@@ -0,0 +1,76 @@
...
+import 'react-image-crop/dist/ReactCrop.css';
+import { Box, Button, Modal, Typography } from '@mui/material';
+import { useState } from 'react';
+import { PixelCrop } from 'react-image-crop';
+
+import getCroppedImg from '../../../../lib/utils/CropperHelper';
+import { styles } from './CropperModal.styles';
+import { ImageCropper } from './ImageCropper/ImageCropper';
+
+interface CropperModalPrpos {
```
uliaescha: there is small mistype Prpos -> Props

app/lib/utils/CropperHelper.ts
```diff
@@ -0,0 +1,48 @@
+export default async function getCroppedImg(
+  imageSrc: string,
+  crop: any,
```
uliaescha: maybe it is better to use explicit type PixelCrop from 'react-image-crop' instead of any

app/shared/components/design-system/cropper-modal/CropperModal.tsx
lizabre: didn't notice, thanks!!

app/lib/utils/CropperHelper.ts
lizabre: Yes, you're right, it is clearer now

uliaescha (approved):

yur4uwe (approved):

VKormylo (approved):

PR #110, opened 2025-07-16
Implement button component

### LINKED ISSUE #95: (SP: 0,5) Move Button Component and Styling from Client to Shared for Reuse in Admin Panel
Description:
We need to move the Button component from the client repository to the admin repository to enable its use within the admin panel UI. The main focus of this task is to:
**Define which button variants (e.g. primary, secondary etc.) are needed in the admin panel**
Example:
Image

uliaescha:
### Moved button component from client repository. 

props you need to use if you want to change styles: 
- variant: _filled, outined, text_
- size: _small, medium, large_
- color: _primary, secondary, tertiary_
- you can also provide logic to be executed when the button is clicked by passing it to the `onClick` prop

`<Button label="Скасувати зміни" color="primary" size="medium" variant="outlined" />`

**Expamles of buttons:** 

image_2025-07-16_21-55-59

### DIFF
--- a/app/shared/components/design-system/button/Button.test.tsx
+++ b/app/shared/components/design-system/button/Button.test.tsx
@@ -8,7 +8,7 @@ describe('Button Component', () => {
   const startIcon = <span data-testid="start-icon">▲</span>;
   const endIcon = <span data-testid="end-icon">▼</span>;
 
-  test('displays icons when provided', () => {
+  test('should display icons when provided', () => {
     render(
       <Button startIcon={startIcon} endIcon={endIcon}>
         Icons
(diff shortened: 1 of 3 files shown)


app/shared/components/design-system/button/Button.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+import React from 'react';
+
+import Button from './Button';
+import { colors } from './Button.styles';
+
+describe('Button Component', () => {
+  const startIcon = <span data-testid="start-icon">▲</span>;
+  const endIcon = <span data-testid="end-icon">▼</span>;
+
+  test('displays icons when provided', () => {
```
oleg191006: I think in the admin we also write tests through IT and names starting with should

uliaescha: yes, sure, fixed that

oleg191006 (approved):

Iarynovskyi (approved):

VKormylo (approved):

danikua (approved):

PR #117, opened 2025-07-17
Added discard changes modal

### LINKED ISSUE #77: (SP: 3) Implement "Cancel Changes" Confirmation Modal
**Objective**
Design and implement a confirmation modal that warns the user before discarding unsaved changes. The modal serves as a safeguard to prevent accidental data loss when navigating away from a form or editable state.

**Modal Content**
The modal should display the following text in Ukrainian:

- Title:
 Скасувати зміни?

- Description:
Усі внесені зміни буде втрачено. Ви впевнені, що хочете вийти без збереження?

Buttons
The modal must include two buttons with the following labels and functions:

Повернутись – Closes the modal and allows the user to stay on the current page (dismiss action).

Скасувати зміни – Confirms the discard action.

For now, only the UI should be implemented. The modal does not need to be connected to any specific form or logic yet, as the exact usage context is still to be defined.
**Reference:**
Image

Renatavl:
## Description

Implemented discard changes modal

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## Video / Screenshots

image_2025-07-17_23-35-56

## Checklist

- [x] Created modal component with proper structure and layout
- [x] Title and description display the correct text
- [x] Added 2 action buttons
- [x] Modal component is reusable

---

### DIFF
--- a/app/shared/components/design-system/button/Button.tsx
+++ b/app/shared/components/design-system/button/Button.tsx
@@ -60,14 +60,13 @@ const Button = forwardRef(
 
     return (
       <MuiButton
-        sx={Object.assign(
-          {},
-          buttonBaseStyles,
-          sizeStyles[size],
-          typographyStyles[color]?.[size],
-          color === 'tertiary' ? variantStyles.tertiary.filled : variantStyles[color]?.[variant],
-          sx
-        )}
+        sx={{
+          ...buttonBaseStyles,
+          ...sizeStyles[size],
+          ...typographyStyles[color]?.[size],
+          ...(color === 'tertiary' ? variantStyles.tertiary.filled : variantStyles[color]?.[variant]),
+          ...(sx as object)
+        }}
         disabled={isDisabled}
         ref={forwardedRef}
         {...props}
--- a/app/shared/components/design-system/button/Button.tsx
+++ b/app/shared/components/design-system/button/Button.tsx
@@ -33,6 +33,7 @@ type Ref = MuiButtonProps['ref'];
 const Button = forwardRef(
   (
     {
+      sx,
       size = 'medium',
       variant = 'filled',
       color = 'secondary',
@@ -59,12 +60,14 @@ const Button = forwardRef(
 
     return (
       <MuiButton
-        sx={{
-          ...buttonBaseStyles,
-          ...sizeStyles[size],
-          ...typographyStyles[color]?.[size],
-          ...(color === 'tertiary' ? variantStyles.tertiary.filled : variantStyles[color]?.[variant])
-        }}
+        sx={Object.assign(
+          {},
+          buttonBaseStyles,
+          sizeStyles[size],
+          typographyStyles[color]?.[size],
+          color === 'tertiary' ? variantStyles.tertiary.filled : variantStyles[color]?.[variant],
+          sx
+        )}
         disabled={isDisabled}
         ref={forwardedRef}
         {...props}
--- a/app/shared/components/design-system/discard-changes-modal/DiscardChangesModal.test.tsx
+++ b/app/shared/components/design-system/discard-changes-modal/DiscardChangesModal.test.tsx
@@ -1,0 +1,50 @@
+import { fireEvent, render, screen } from '@testing-library/react';
+
+import DiscardChangesModal from './DiscardChangesModal';
+
+describe('DiscardChangesModal', () => {
+  const setup = (overrideProps = {}) => {
+    const defaultProps = {
+      open: true,
+      handleClose: jest.fn(),
+      handleSubmit: jest.fn()
+    };
+
+    const props = { ...defaultProps, ...overrideProps };
+    render(<DiscardChangesModal {...props} />);
+    return props;
+  };
+
+  it('renders modal with title and description', () => {
+    setup();
+
+    expect(screen.getByText('Скасувати зміни?')).toBeInTheDocument();
+    expect(
+      screen.getByText('Усі внесені зміни буде втрачено. Ви впевнені, що хочете вийти без збереження?')
+    ).toBeInTheDocument();
+  });
+
+  it('calls handleClose when "Повернутись" button is clicked', () => {
+    const { handleClose } = setup();
+
+    const closeButton = screen.getByRole('button', { name: /Повернутись/i });
+    fireEvent.click(closeButton);
+
+    expect(handleClose).toHaveBeenCalled();
+  });
+
+  it('calls handleSubmit when "Скасувати зміни" button is clicked', () => {
+    const { handleSubmit } = setup();
+
+    const submitButton = screen.getByRole('button', { name: /Скасувати зміни/i });
+    fireEvent.click(submitButton);
+
+    expect(handleSubmit).toHaveBeenCalled();
+  });
+
+  it('does not render content when modal is closed', () => {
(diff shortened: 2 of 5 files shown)


app/shared/components/design-system/discard-changes-modal/DiscardChangesModal.tsx
```diff
@@ -0,0 +1,40 @@
+'use client';
+
+import { Box, Button, Modal, Typography } from '@mui/material';
```
Iarynovskyi: Maybe it would be better to use custom button from ds-system ?

app/shared/components/design-system/discard-changes-modal/DiscardChangesModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { styles } from './DiscardChangesModal.styles';
+
+interface DiscardChangesModalProps {
+  open: boolean;
+  handleClose: () => void;
+  handleSubmit: () => void;
+}
+
+const DiscardsChangesModal: React.FC<DiscardChangesModalProps> = ({ open, handleClose, handleSubmit }) => {
```
Iarynovskyi:
```suggestion
const DiscardChangesModal: React.FC<DiscardChangesModalProps> = ({ open, handleClose, handleSubmit }) => {
```
Because you write it in filename without 's'

Iarynovskyi: And write tests for this component, please

app/shared/components/design-system/discard-changes-modal/DiscardChangesModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+  return (
+    <Modal sx={styles.modal} open={open} onClose={handleClose}>
+      <Box sx={styles.contentWrapper}>
+        <Box sx={styles.modalHeader}>
+          <Image src="./icons/warning.svg" alt="warning" width={40} height={40} />
```
Iarynovskyi:
```suggestion
          <Image src="/icons/warning.svg" alt="warning" width={40} height={40} />
```

Renatavl:
@Iarynovskyi done

app/shared/components/design-system/button/Button.tsx
```diff
@@ -59,12 +60,14 @@ const Button = forwardRef(
...
+        sx={Object.assign(
+          {},
+          buttonBaseStyles,
+          sizeStyles[size],
+          typographyStyles[color]?.[size],
+          color === 'tertiary' ? variantStyles.tertiary.filled : variantStyles[color]?.[variant],
+          sx
+        )}
         disabled={isDisabled}
         ref={forwardedRef}
```
oleg191006, line 71: you don't need to add sx here, it will work without it, because below we have ...props which will include sx, and it is also better to use the spread operator instead of Object.assign, so this component can be left unchanged at all

app/shared/components/design-system/discard-changes-modal/DiscardChangesModal.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+      handleClose: jest.fn(),
+      handleSubmit: jest.fn()
+    };
+
+    const props = { ...defaultProps, ...overrideProps };
+    render(<DiscardChangesModal {...props} />);
+    return props;
+  };
+
+  it('renders modal with title and description', () => {
```
oleg191006: all test names start with should

luvthenika (approved):

app/shared/components/design-system/button/Button.tsx
VKormylo: It will work without it, but it will overwrite all styles for button, since we overwrite applied sx with passed sx prop. So this approach is correct in my opinion, because we want to extend styles and not overwrite completely. Working on a task and had this problem. And I agree, spread looks cleaner.

Iarynovskyi (approved):
Looks good to me, but do not forget to fix Sonar issue

app/shared/components/design-system/button/Button.tsx
oleg191006: > It will work without it, but it will overwrite all styles for button, since we overwrite applied sx with passed sx prop. So this approach is correct in my opinion, because we want to extend styles and not overwrite completely. Working on a task and had this problem.

hmm, that makes sense, then just replace Object.assign with spread operator

Renatavl:
@oleg191006 I tried to use "...sx" but unfortunately it doesn't work because of type issues (see the screenshot). So I found a solution using Object.assign and it works without errors. Maybe you have any ideas about how to fix the type error?

image_2025-07-18_16-15-49

VKormylo:
> @oleg191006 I tried to use "...sx" but unfortunately it doesn't work because of type issues (see the screenshot). So I found a solution using Object.assign and it works without errors. Maybe you have any ideas about how to fix the type error?
> image_2025-07-18_16-15-49

You can type all of styles except sx prop like this: `as SxProps`, it should help, or just use .assign there in order to avoid these type assertions, up to you.

oleg191006:
> @oleg191006 I tried to use "...sx" but unfortunately it doesn't work because of type issues (see the screenshot). So I found a solution using Object.assign and it works without errors. Maybe you have any ideas about how to fix the type error?
> 
> image_2025-07-18_16-15-49

 sx={{
          ...(buttonBaseStyles),
          ...(sizeStyles[size] ),
          ...(typographyStyles[color]?.[size]),
          ...((color === 'tertiary' ? variantStyles.tertiary.filled : variantStyles[color]?.[variant]) ),
          ...(sx as SxProps)
        }} this should help

Renatavl:
> sx={{ ...(buttonBaseStyles), ...(sizeStyles[size] ), ...(typographyStyles[color]?.[size]), ...((color === 'tertiary' ? variantStyles.tertiary.filled : variantStyles[color]?.[variant]) ), ...(sx as SxProps) }} this should help

Tried both approaches and none of them works for me

image

Renatavl:
> > sx={{ ...(buttonBaseStyles), ...(sizeStyles[size] ), ...(typographyStyles[color]?.[size]), ...((color === 'tertiary' ? variantStyles.tertiary.filled : variantStyles[color]?.[variant]) ), ...(sx as SxProps) }} this should help
> 
> Tried both approaches and none of them works for me

Tried this approach, should I leave it?

image

oleg191006:
> > sx={{ ...(buttonBaseStyles), ...(sizeStyles[size] ), ...(typographyStyles[color]?.[size]), ...((color === 'tertiary' ? variantStyles.tertiary.filled : variantStyles[color]?.[variant]) ), ...(sx as SxProps) }} this should help
> 
> Tried both approaches and none of them works for me
> 
> image

so for me the error disappeared at first, and then appeared again, sx={{
...(buttonBaseStyles as SxProps),
...(sizeStyles[size] as SxProps),
...(typographyStyles[color]?.[size] as SxProps),
...(color === 'tertiary' ? variantStyles.tertiary.filled : (variantStyles[color]?.[variant] as SxProps)),
...(sx as SxProps)
}}  
so it doesn't give an error with this entry

danikua (approved):

PR #195, opened 2025-08-13
 (SP: 3) Integrate Zustand slice with editable text fields and track unsaved changes flag

### LINKED ISSUE #181: (SP: 3) Integrate Zustand slice with editable text fields and track unsaved changes flag
## Description:

- Replace local React useState in components (e.g., LiatoshynskyFoundation, LiatoshynskyOffice, QuoteBlock, etc.) with state management via a Zustand slice
- For each editable field (text, paragraphs, quotes), create corresponding actions in the Zustand slice to update specific blocks or fields
- On the first change of any field, set the `isChanged` flag in the Zustand slice to true
- Ensure all fields receive their current values from Zustand state instead of local component state
- Support editing complex structures, such as arrays of paragraphs, allowing updates to individual items
- Make sure the UI reacts to Zustand state changes properly
- Avoid duplicating state locally; all changes and reads should go through Zustand

### Requirements:
- Change input fields from uncontrolled (defaultValue) to controlled (value) components with values from Zustand
- Dispatch Zustand actions on each field change to update state
- In the update action, set isChanged = true if it’s not already set

### LINKED ISSUE #185: (SP: 1) Implement “Save” with publish and draft functionality
## Description:

- Add a Save button to the page editor that saves the current content to the database with status: published or draft

### In the scope of this task, create separate Zustand actions for:

- saveAsDraft – saves the current page as draft
- publishPage – saves the current page as published

### Both actions should:

- Collect the current page content from the store
- Reset the isChanged flag after successful save

oleg191006:
## Description
Created createEditSlice , a universal Zustand slice for managing editable page blocks.
Added useInitBlock , a generic hook for initializing and retrieving block data from the global store. Ensures that the block is populated with initial data if it's not yet present
Added:

- setField — updates a single field in a block if value changed.
- setFields — updates multiple fields with initialization support (isInit flag)
- introduced isChanged and isInitialized flags for tracking edit state and initialization

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1.Use any block that uses global state
2.Add to this comoponent
```
 const blocks = useStore((state) => state.blocks) || {};
  const isChanged = useStore((state) => state.isChanged);
  console.log(blocks);
  console.log(isChanged);
```

## Video / Screenshots

https://github.com/user-attachments/assets/8960b305-5352-4ae8-98cc-9afa781b27d2

https://github.com/user-attachments/assets/13c00d5a-1f7e-435a-be4e-a185d314289a

## Checklist

- [x] Tests pass successfully
- [ ] No Sonar issues

---

### DIFF
--- a/app/shared/components/about-us/Entry-section/EntrySection.tsx
+++ b/app/shared/components/about-us/Entry-section/EntrySection.tsx
@@ -6,12 +6,13 @@ import { ImagePreviewBlock } from '../../design-system/photo-block/PhotoBlock';
 import { CustomTextField } from '../../design-system/text-field/TextField';
 import { QuoteBlock } from '../Liatoshynsky-office/quote-block/QuoteBlock';
 import { hardcodedData } from './EntrySection.consts';
+import { BLOCK_IDS, PAGE_IDS } from '~/constants/pageBlocks';
 import useInitBlock from '~/shared/hooks/use-init-block/useInitBlock';
 import { useStore } from '~/store';
 
 export const EntrySection = () => {
-  const pageId = 'aboutUs';
-  const blockId = 'entrySection';
+  const pageId = PAGE_IDS.ABOUT_US;
+  const blockId = BLOCK_IDS.ENTRY_SECTION;
 
   const block = useInitBlock(pageId, blockId, hardcodedData);
 
--- a/app/shared/components/about-us/Entry-section/EntrySection.tsx
+++ b/app/shared/components/about-us/Entry-section/EntrySection.tsx
@@ -1,19 +1,21 @@
 'use client';
 import { Box } from '@mui/material';
-import { useState } from 'react';
 
 import CollapsibleBlock from '../../design-system/collapsible-block/CollapsibleBlock';
 import { ImagePreviewBlock } from '../../design-system/photo-block/PhotoBlock';
 import { CustomTextField } from '../../design-system/text-field/TextField';
 import { QuoteBlock } from '../Liatoshynsky-office/quote-block/QuoteBlock';
 import { hardcodedData } from './EntrySection.consts';
+import useInitBlock from '~/shared/hooks/use-init-block/useInitBlock';
+import { useStore } from '~/store';
 
 export const EntrySection = () => {
-  const [title, setTitle] = useState(hardcodedData.title);
-  const [image, setImage] = useState(hardcodedData.image);
-  const [imageCaption, setImageCaption] = useState(hardcodedData.imageCaption);
-  const [quoteText, setQuoteText] = useState(hardcodedData.quoteText);
-  const [quoteDescription, setQuoteDescription] = useState(hardcodedData.quoteDescription);
+  const pageId = 'aboutUs';
+  const blockId = 'entrySection';
+
+  const block = useInitBlock(pageId, blockId, hardcodedData);
+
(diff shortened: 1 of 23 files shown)


ssashayurchenko (commented):
Maybe it would be better to add debounce logic to prevent excessive requests and frequent Zustand state updates, since it currently reacts to every user keystroke in the input fields.

oleg191006:
> Maybe it would be better to add debounce logic to prevent excessive requests and frequent Zustand state updates, since it currently reacts to every user keystroke in the input fields.

well, when we make changes to the fields, we don't send requests at all, the request may be in the future when we click the save button to save to the database. Right now there are no heavy calculations or network calls, so I think debounce is not needed here

app/shared/components/about-us/Entry-section/EntrySection.tsx
```diff
@@ -1,52 +1,56 @@
...
+import { useStore } from '~/store';
 
 export const EntrySection = () => {
-  const [title, setTitle] = useState(hardcodedData.title);
-  const [image, setImage] = useState(hardcodedData.image);
-  const [imageCaption, setImageCaption] = useState(hardcodedData.imageCaption);
-  const [quoteText, setQuoteText] = useState(hardcodedData.quoteText);
-  const [quoteDescription, setQuoteDescription] = useState(hardcodedData.quoteDescription);
+  const pageId = 'aboutUs';
+  const blockId = 'entrySection';
```
Iarynovskyi: I think all IDs we need to save in one place to reuse it and faster change in future 
Because now you use them in many files

app/store/slices/editSlice.ts
```diff
@@ -0,0 +1,63 @@
...
+  saveAsDraft: (pageId: string) => {
+    const _pageBlocks = get().blocks[pageId];
+
+    set({ isChanged: false });
+  },
+  publishPage: (pageId: string) => {
+    const _pageBlocks = get().blocks[pageId];
+
+    set({ isChanged: false });
+  }
```
Iarynovskyi, line 62: Can we combine them into one ?
And just put draft or published as a parameter inside, because logic are same

oleg191006: I noticed that the logic is currently the same, but I left them separate because in the future they will probably have different API requests or additional logic for draft or publish. What do you think about this?

app/shared/components/about-us/Entry-section/EntrySection.tsx
oleg191006: yes I completely agree, I'll change it now

app/store/slices/editSlice.ts
Iarynovskyi: I think that the saving logic will be making by the one resolver, but you are right that schemas for it can be little different 

So for now let it be 
Maybe in future it will be changed by someone, but for now it's okay

Iarynovskyi (approved):

danikua (approved):

Kryzhanivsky (approved):
approved, but there are still 2 sonar issues left

oleg191006:
> approved, but there are still 2 sonar issues left

I saw it, but since the typing is complicated here, there aren't many options

PR #214, opened 2025-08-26
Feature/182/dynamic page content

### LINKED ISSUE #182: (SP: 5) Implement backend logic to fetch page content from db and serve it to admin panel dynamically
## Description:

- Replace hardcoded static data currently used in page content blocks with dynamic data fetched from the database
- Implement a repository layer responsible for querying the database and returning the content blocks for the specific page
- Create a service layer that uses the repository to fetch and, if necessary, transform or aggregate the data before sending it to the client
- Develop GraphQL resolvers to expose the service data through the API, allowing the client to query page content dynamically
- Ensure the resolver schema supports the structure of content blocks as expected by frontend components
- On the frontend, refactor page components and content blocks to fetch their data from the GraphQL API instead of using hardcoded data

###  Requirements:

- **Repository:** implement methods to retrieve page content blocks from the database by page identifier or slug
- **Service:** implement business logic, caching (if needed), and data preparation
- **GraphQL:** define schema types and queries as needed, implement resolvers
- **Frontend:** replace static imports/data with data fetched via GraphQL queries, integrate with Zustand or other state management if necessary

oleg191006:
## Description
Replaced hardcoded data with page blocks from the DB, added usePageBlock hook to fetch and return block by pageId and blockId, extended global state with setPageData and locale management functions

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Ensure a database is connected and accessible
2. Navigate to the About-us page and verify that content is loaded from the database instead of hardcoded values

## Video / Screenshots

https://github.com/user-attachments/assets/8e0ed5d7-dfa7-4d83-aba3-65c2febb63ec

## Checklist

- [x] Tests pass successfully
- [ ] No Sonar issues

---


app/shared/components/about-us/Entry-section/EntrySection.tsx
Iarynovskyi, line 1: Maybe it would be better to rename this file and component inside to IntroSection too ?

app/shared/components/about-us/Entry-section/EntrySection.tsx
```diff
@@ -1,59 +1,93 @@
...
-        <QuoteBlock
-          title={block.quoteText || ''}
-          description={block.quoteDescription || ''}
-          onTitleChange={(val) => setField(pageId, blockId, 'quoteText', val)}
-          onDescriptionChange={(val) => setField(pageId, blockId, 'quoteDescription', val)}
-        />
-      </Box>
-    </CollapsibleBlock>
+    <>
+      {block && (
```
Iarynovskyi: Its unnecessary check because you checked if block exist before in 23 line

app/shared/components/about-us/Liatoshynsky-foundation/LiatoshynskyFoundation.tsx
```diff
@@ -1,49 +1,74 @@
...
   return (
     <CollapsibleBlock title="Фундація Лятошинського">
       <FoundationBlock
-        mainText={block.mainText || ''}
-        paragraphs={block.paragraphs || []}
-        imageUrl={`/images/${block.image || ''}`}
-        fileName={block.imageFileName}
+        mainText={mainText}
+        paragraphs={paragraphs}
+        imageUrl={`/images/${block.image?.src}.png`}
```
Iarynovskyi, line 68: For now let it be, but in future when we will start using blob storage we will need to refactor this logic, because photo will be getting from storage instead of files inside project

app/lib/utils/prose.ts
Iarynovskyi, line 1: I know why you created this util, and i see problem that i think you know about too
The `content` here is TipTap document and it can have styles for text
You get from `context` only text and forgot about styles, so than we in input see simple text -> than when you use `textToProse` you will totally lost information about styles
`
type: "text",
text: "Фундація Лятошинського",
marks: [
type: {
"bold"
}
]
`
for example this `blod` style

But for my opinion its okay here, and maybe during current sprint, we will create some `TextEditor` that will work with TipTap data correctly, because its not part of your work here

app/shared/components/about-us/our-mission/OurMission.tsx
```diff
@@ -1,166 +1,194 @@
...
+      ...(file && { src: file.name, generatedSrc: `/api/blob-url?folderName=photos&blobName=${file.name}` })
     };
-    const updated = [...block.missionPoints, newPoint];
-    setField(pageId, blockId, 'missionPoints', updated);
-    return newPoint;
+
+    setField(pageId, blockId, key, updated);
+  };
+
+  const handleImageFileChange = (key: 'smallImage' | 'bigImage', file: File) => {
```
Iarynovskyi, line 119: app/shared/components/about-us/Liatoshynsky-foundation/LiatoshynskyFoundation.tsx

In this file you have very similar `handleImageChange` so maybe we can move this logic to some separate util ?

src/infrastructure/repositories/pageRepository/pageRepository.ts
```diff
@@ -0,0 +1,25 @@
+import mongoose from 'mongoose';
+
+import { BasePage } from '~/domain/entities/Page';
+import dbConnect from '~/infrastructure/db/connect';
+
+export const PageRepository = () => ({
+  getPageBySlug: async (slug: string): Promise<BasePage | null> => {
+    await dbConnect();
+    const pageCollection = mongoose.connection.collection('pages');
```
Iarynovskyi: Move Page schema from client repo and use it here instead of `mongoose.connection.collection('smth')`

src/interfaces/graphql/resolvers/admin/Query.ts
```diff
@@ -12,5 +14,15 @@ export const Query = {
...
+  },
+  pageBlocks: async (_: unknown, args: { slug: string }) => {
+    const service = PageService(PageRepository());
+    const page = await service.getPage(args.slug);
+
+    if (!page) {
+      throw new GraphQLError('Page not found', { extensions: { code: 'NOT_FOUND' } });
+    }
+
+    return page;
```
Iarynovskyi:
```suggestion
  pageBlocks: async (_: unknown, args: { slug: string }) => {
    if (!context.admin) {
      throw new GraphQLError('You must be logged in to access this resource.', {
        extensions: {
          code: 'UNAUTHENTICATED'
        }
      });
    }
    const service = PageService(PageRepository());
    const page = await service.getPage(args.slug);

    if (!page) {
      throw new GraphQLError('Page not found', { extensions: { code: 'NOT_FOUND' } });
    }

    return page;
```

add it here please to handle UNAUTHENTICATED admins

app/shared/components/about-us/Entry-section/EntrySection.tsx
oleg191006: yes indeed, I forgot to take it, thank you

app/lib/utils/prose.ts
oleg191006: I also thought it would be better as a separate task, since there are already so many changes

app/shared/components/about-us/our-mission/OurMission.tsx
oleg191006: The idea is cool, but there is a caveat here: to implement this in the utility, you will have to pass a bunch of arguments (pageId, blockId, setField, the block itself and the key), and because of this, the code becomes more cumbersome. Plus, there is also the hassle with types, because TS starts to be very picky about matching keys and block fields, so as a result, the utility does not simplify much, but rather adds unnecessary complexity

Iarynovskyi: okay, i got it

Iarynovskyi (approved):

app/shared/components/about-us/our-mission/OurMission.tsx
```diff
@@ -1,166 +1,194 @@
...
-      id: crypto.randomUUID(),
-      value: ''
+  const handleImageChange = (key: 'smallImage' | 'bigImage', value: string, file?: File) => {
+    const image = block[key];
+    if (!image) return;
+
+    const updated = {
+      ...image,
+      caption: { ...image.caption, [currentLocale]: value },
+      ...(file && { src: file.name, generatedSrc: `/api/blob-url?folderName=photos&blobName=${file.name}` })
```
ssashayurchenko, line 113: maybe it would be better to move this path into constants?
```suggestion
      ...(file && { src: file.name, generatedSrc: `${BLOB_API_PATH}?folderName=photos&blobName=${file.name}` })
```

danikua (approved):

app/shared/components/about-us/our-mission/OurMission.tsx
```diff
@@ -1,166 +1,194 @@
...
   return (
-    <CollapsibleBlock title={OUR_MISSION_TEXT.sectionTitle}>
+    <CollapsibleBlock title="Наша місія">
       <Box sx={styles.wrapper}>
         <CustomTextField
-          title={OUR_MISSION_TEXT.titleFieldTitle}
-          label={OUR_MISSION_TEXT.titleFieldLabel}
-          value={block.title || ''}
+          title="Заголовок секції"
+          label="Текст заголовка"
```
Kryzhanivsky, line 137: Can we replace these hardcoded lines, or at least use a translation? Same in other places

Kryzhanivsky (approved):

app/shared/components/about-us/our-mission/OurMission.tsx
oleg191006: yes it is possible, but for now it is just a stub, the final functionality will not look like this. I think whoever is going to do the blob will decide how best to do it

app/shared/components/about-us/our-mission/OurMission.tsx
oleg191006: well, as far as I remember, we agreed that everything in our admin area would be in Ukrainian, right?

ssashayurchenko (approved):

PR #318, opened 2025-11-17
fix: fix login bug when user press enter

ruslansymonenko:
## Description

Admin login form did not submit when Enter key is pressed after entering credentials. Fixed this bug, and update tests.

## Type of change
- [x] Bug fix

## How to test

1. Run app in dev mode
2. Make sure you are connected to the database.
3. Try logging into the website by clicking the button "Login", and also with pressing Enter. Both methods should work.

---


yur4uwe (approved):
You see this wasn't a bug, simply intended functionality, though bit incorrect. Anyway thanks for your work)

Taras-ep (approved):

stsvt (approved):

PR #328, opened 2025-11-21
feat: create universal slug generator

### LINKED ISSUE #945: (SP: 1)  Implement universal slug generation functionality
## Description

Add a reusable system for automatically generating slugs from entity titles to ensure consistent and SEO-friendly URLs across the project.

ruslansymonenko:
## Description

Add reusable slug generation utility with automatic uniqueness check

## Type of change

- [x] New feature

### DIFF
--- a/src/constants/errors.ts
+++ b/src/constants/errors.ts
@@ -17,5 +17,5 @@ export const utilsErrors = {
   EMPTY_TITLE_FOR_SLUG: 'Title must be a non-empty string',
   NO_BASE_SLUG_AVAILABLE: 'Title produces an empty slug and no fallback slug is provided',
   EMPTY_SLUG: 'Normalized slug is empty',
-  UNABLE_GENERATE_UNIQUE_SLUG: 'Unable to generate unique slug after max attempts'
+  SLUG_GENERATION_ERROR: 'Unable to generate unique slug after max attempts'
 };
--- a/src/constants/errors.ts
+++ b/src/constants/errors.ts
@@ -12,3 +12,10 @@ export const errors = {
   INVALID_EMAIL_FORMAT: 'Invalid email format',
   EMAIL_TOO_LONG: 'Email must not exceed 254 characters'
 };
+
+export const utilsErrors = {
+  EMPTY_TITLE_FOR_SLUG: 'Title must be a non-empty string',
+  NO_BASE_SLUG_AVAILABLE: 'Title produces an empty slug and no fallback slug is provided',
+  EMPTY_SLUG: 'Normalized slug is empty',
+  UNABLE_GENERATE_UNIQUE_SLUG: 'Unable to generate unique slug after max attempts'
+};
--- a/src/shared/utils/slugGenerator/slugGenerator.ts
+++ b/src/shared/utils/slugGenerator/slugGenerator.ts
@@ -25,7 +25,9 @@ type SlugGeneratorOptions = {
 };
 
 function normalizeSlug(slug: string): string {
-  return slug.replace(/-+/g, '-').replace(/^-+|-+$/g, '');
+  const replaced = slug.replace(/-+/g, '-');
+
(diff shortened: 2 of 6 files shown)


src/constants/errors.ts
```diff
@@ -12,3 +12,10 @@ export const errors = {
   INVALID_EMAIL_FORMAT: 'Invalid email format',
   EMAIL_TOO_LONG: 'Email must not exceed 254 characters'
 };
+
+export const utilsErrors = {
+  EMPTY_TITLE_FOR_SLUG: 'Title must be a non-empty string',
+  NO_BASE_SLUG_AVAILABLE: 'Title produces an empty slug and no fallback slug is provided',
+  EMPTY_SLUG: 'Normalized slug is empty',
+  UNABLE_GENERATE_UNIQUE_SLUG: 'Unable to generate unique slug after max attempts'
```
yur4uwe: Not to be very pesky but in my opinion `SLUG_GENERATION_ERROR` has a better ring

src/shared/utils/slugGenerator/slugGenerator.ts
```diff
@@ -0,0 +1,94 @@
...
+    const uniqueSlug = normalizeSlug(`${baseSlug}-${counter}`);
+
+    const slugExists = await checkExists(uniqueSlug);
+    if (!slugExists) {
+      return uniqueSlug;
+    }
+
+    counter++;
+    attempts++;
+  }
```
yur4uwe, line 91: sorry that i don't remember, but why didn't we used unix timestamp to avoid name collisions?

ruslansymonenko: Because using a readable slug (regarding best practice) makes the URL understandable for users, improves SEO (since Google prefers URLs that reflect the product/item name), and makes links easier to remember. A timestamp would solve collisions, but it would remove all these benefits.

alisa-korniienko (approved):

src/shared/utils/slugGenerator/slugGenerator.ts
yur4uwe: Yes, but i think you can replace only counter with a unix timestamp and have still readable slug and wouldn't need to calculate it with expensive `checkExists`

ruslansymonenko: I agree that using a timestamp is simpler and more convenient—I would personally prefer that approach.
However, the decision isn’t really up to me, it depends on what the client needs.
I’m not sure how it impacts SEO, so it’s hard for me to say which approach is “better.”
I found information that the counter approach is popular for small systems.
Since the slug is primarily an SEO question (because without slug we can just use ID only), I think we should discuss it in the meeting with experts.

yur4uwe: I agree, but we should anyway look into how it would impact SEO ourselves to not come to the meeting empty handed

ruslansymonenko: I looked through different approaches for generating slugs, and we discussed them in the meeting. It was decided to stick with the counter.

For SEO:
The slug stays short, readable, and reflects the title.
People and search engines can more easily remember and index clear URLs.
Unlike a Unix timestamp, the URL doesn’t contain long numbers that carry no meaningful information.

For small systems (like we have):
The counter works well because the system doesn’t have a high volume of record creation (news and event will be created only by admin), which makes the possibility of conflicts minimal.

Additionally, our items depend on time (news date, event date). A potential issue with search engine indexing is that the slug would contain the creation date of the news, while the client requires the ability to set and manage the news date themselves. This can result in a discrepancy between the date in the slug and the actual date set by the client.

yur4uwe: Then so be it

yur4uwe (approved):

kolibri753 (approved):

PR #439, opened 2026-04-03
Create publications page with header and publication cards grid

### LINKED ISSUE #394: (SP: 3) Create publications(news and events) page
## **Description:** 
Build the main `/publications` dashboard page. It requires a functional header for controls (search/filter/create) and a grid of interactive cards displaying News, Events, and Media Mentions.

## **Technical Requirements:**

- **1. Publications Header Component**
    - Properly use the reusable header with search and filters component.
        
- **2. Publication Card Component**
    - Build or use a reusable MUI card that displays main polymorphic data (Publication type, Publication status, Internal title, Created at date.
    - Display core entity information inside the card.
    - Implement an "Edit" action/button on the card that routes to `/publications/[type]/[id]/edit`.
        
- **3. Main Dashboard Layout**
    - Implement a grid system below the header to render the fetched list of publication cards.

## **Expected result:**

Image

## **Extra:**
You definitely should refer to [Figma design](https://www.figma.com/design/IDQ5XmNo6PQBmBlL3DBm6Q/Admin-Liatoshynsky-Foundation?node-id=4781-74266&t=qWbap0whtcXMI7J3-0).

kandyba:
## Description

This PR implements the main **Publications (/publications) page** in the admin panel.

The page includes:
- Reusable **Header component** with search and filtering controls
- A grid layout displaying **publication cards** (News, Events, Media Mentions)
- Reusable **Publication Card component** with core entity data

The grid system manages the layout and supports rendering a dynamic list of publication cards.

link:
https://github.com/Liatoshynsky-Foundation/lf-admin/issues/394

## Type of change

- [x] New feature

## How to test

1. Open the **/publications** page.
2. Verify that the **header is displayed** with:
   - Search input
   - Filters
   - Action controls
3. Check the **publications grid**:
   - Cards are displayed correctly
   - Each card shows type, status, title, and created date
4. Click the **Edit button** on any card:
   - Verify navigation to `/publications/[type]/[id]/edit`
5. Verify that:
   - Grid layout works correctly
   - Cards are rendered dynamically
6. Ensure no UI or functional issues occur.

## Video / Screenshots

Screenshot 2026-04-03 at 13 04 02
---


app/shared/components/selector/FilterSelect.styles.ts
dest411, line 89: Awesome PR! Btw, I noticed that long texts (like "Опублікована з чернеткою") squish this delete icon.
Adding `flexShrink: 0` right here fixes it perfectly.

video:
https://github.com/user-attachments/assets/fdbfd315-9cbb-4aa0-8015-f59ed7b9f42e

dest411 (approved):

Th0mas-H0ward (changes requested):
Great work! However, I've just found that grid items not spanning full width - the grid leaves a significant gap on the right side, as you can see on a screenshot (this issue was found on Windows, in Google Chrome Version 146.0.7680.153 (Official Build) (64-bit); device resolution: `1920x1080`; browser viewport: `1536x695`).

Image

navimov (approved):

app/shared/hooks/use-publications/usePublicationsFiltering.ts
```diff
@@ -0,0 +1,200 @@
...
+      onValueChange: handleSortValueChange
+    }),
+    [currentSortField, currentSortOption.label, handleSortFieldChange, handleSortValueChange, sortValue]
+  );
+
+  return {
+    filteredItems,
+    toolbarProps,
+    sortProps
+  };
```
Fedorieieva: The current implementation of usePublicationsFiltering performs filtering and sorting in-memory on the client side using useMemo, but we should move this logic to the backend to ensure scalability. Our BaseRepository and specific repositories like NewsRepository or MediaMentionsRepository already support FiltersInput and SortCriteria at the database level. Please refactor this hook to pass these parameters to the API, allowing the server to handle the heavy lifting as we've already implemented in our backend infrastructure

Fedorieieva (changes requested):

app/(logged_in)/publications/PublicationsPageContent.tsx
```diff
@@ -0,0 +1,629 @@
...
+    uk?: string;
+    en?: string;
+  };
+  coverImage: {
+    src: string;
+    alt: {
+      uk?: string;
+      en?: string;
+    };
+  };
```
Fedorieieva: We have an ImageBlock interface in app/types/common.ts

app/(logged_in)/publications/PublicationsPageContent.tsx
```diff
@@ -0,0 +1,629 @@
...
+  events: T;
+};
+type LocalizedValue = {
+  uk?: string | null;
+  en?: string | null;
+};
+type LocalizedCardValue = {
+  uk?: string;
+  en?: string;
+};
```
Fedorieieva: Same goes for localized values, we have a LocalizedString interface in app/types/common.ts

Th0mas-H0ward (approved):

app/(logged_in)/publications/PublicationsPageContent.styles.ts
```diff
@@ -0,0 +1,28 @@
...
+      md: '0 0 calc((100% - 32px) / 3)',
+      xl: '0 0 calc((100% - 48px) / 4)'
+    },
+    maxWidth: {
+      xs: '100%',
+      sm: 'calc((100% - 16px) / 2)',
+      md: 'calc((100% - 32px) / 3)',
+      xl: 'calc((100% - 48px) / 4)'
+    },
+    justifyContent: 'flex-start'
```
Fedorieieva: We are doing a lot of manual math with calc and gap subtractions here. Have you considered using display: grid for the parent container instead? By using gridTemplateColumns: repeat(N, 1fr) and the gap property, we could eliminate all these complex calculations and the redundancy between flex and maxWidth entirely

app/constants/publications.ts
```diff
@@ -0,0 +1,78 @@
...
+  { value: 'events', label: 'Події', href: '/publications/events' },
+  { value: 'news', label: 'Новини', href: '/publications/news' },
+  { value: 'media', label: 'Ми у ЗМІ', href: '/publications/media' }
+];
+
+export const PUBLICATIONS_CREATE_OPTIONS: ReadonlyArray<PublicationsCreateOption> = [
+  { id: 'event', label: 'Подію', href: '/publications/events/create' },
+  { id: 'news', label: 'Новину', href: '/publications/news/create' },
+  { id: 'media', label: 'Ми у ЗМІ', href: '/publications/media/create' }
+];
```
Fedorieieva: There is quite a bit of redundancy between PUBLICATIONS_TABS and PUBLICATIONS_CREATE_OPTIONS regarding URLs and labels. This makes the code harder to maintain -- for instance, if we decide to change the /publications prefix or update a category name, we'd have to do it in multiple places.

I suggest extracting the common data (slugs and labels) into a base configuration object and then mapping over it to generate these arrays. This would ensure a single source of truth for our publication categories

app/shared/components/content-card/ContentCard.tsx
```diff
@@ -8,24 +10,26 @@ import { getStatus } from '~/lib/utils/getStatus';
 
 export type ContentType = 'news' | 'event' | 'media';
 
+const FALLBACK_IMAGE_SRC = '/images/image.png';
+
+type LocalizedCardValue = {
+  uk?: string;
+  en?: string;
+};
```
Fedorieieva: Same here, this interface duplicates the LocalizedString we already have in app/types/common.ts. Please import it from there instead

app/shared/hooks/use-publications/usePublicationsFiltering.ts
```diff
@@ -0,0 +1,246 @@
...
+  if (status === 'published_with_draft') {
+    return MediaStatus.Editing;
+  }
+
+  if (status === 'published') {
+    return MediaStatus.Published;
+  }
+
+  return MediaStatus.Draft;
+};
```
Fedorieieva: The logic in mapPublicationStatusToNewsStatus and mapPublicationStatusToMediaStatus is identical, with the only difference being the return type (Enum).

We could simplify this by creating a single generic function that accepts the target status Enum as an argument. This would make it much easier to maintain if the backend status logic changes in the future. What do you think about using something like mapPublicationStatus<T>(status, enumObject)?

app/shared/hooks/use-publications/usePublicationsFiltering.ts
```diff
@@ -0,0 +1,246 @@
...
+      { field: 'createdAt', order: SortOrder.Desc }
+    ];
+  }
+
+  if (sortValue === 'date_asc') {
+    return [{ field: 'createdAt', order: SortOrder.Asc }];
+  }
+
+  return [{ field: 'createdAt', order: SortOrder.Desc }];
+};
```
Fedorieieva: The sorting logic for News and Media is identical, but it's currently duplicated across two functions with slightly different implementations (one uses Enums, the other uses strings).

Given that both entities implement BaseContent and share adminTitle and createdAt fields, we could create a single helper function using generics. This would ensure that any changes to our default sorting logic (like the fallback createdAt: Desc) only need to be updated in one place.

Also, I noticed NewsSortBy is used for News, but raw strings are used for Media. Can we unify this to use Enums for both to keep it type-safe?

src/infrastructure/repositories/helpers.ts
```diff
@@ -16,19 +52,63 @@ export const createToEntity = <
...
   }
 
-  return query as FilterQuery<TDb>;
+  return {
+    $and: [
+      ...(hasBaseQuery ? [baseQuery as FilterQuery<TDb>] : []),
+      ...compoundConditions
+    ]
+  } as FilterQuery<TDb>;
 };
```
Fedorieieva, line 98: Regarding the filters type: having both `status?: string` and `statuses?: string[]` feels redundant and potentially confusing. Since `statuses` (as an array) can handle both single and multiple values using the `$in` operator, I suggest removing the single `status` field to have a single source of truth.

And one other oservation: we can simplify the final assembly by pushing all conditions (including `status` and `slug`) into a single `conditions` array. Then, we can return:
    * `{}` if the array is empty.
    * `conditions[0]` if there is only one element.
    * `{ $and: conditions }` if there are multiple.
    This would eliminate the need for `hasBaseQuery` and the complex `if` checks at the end

Fedorieieva (changes requested):

kolibri753 (approved):

Fedorieieva (approved):

Issue #428, opened 2026-03-31
(SP: 3) [Feature] Implement persistent storage for image crop coordinates


Fedorieieva:
## Description
Following the UI fixes in issue #405 and the unification of `MediaModal` in #424, the application now correctly handles crop state in the frontend. However, these coordinates are currently stored only in local component state and are lost upon page refresh. This task involves updating the backend schema and API endpoints to persist crop data (`x`, `y`, `width`, `height`) in the database.

## Problem Statement
Currently, image cropping is "volatile":
- **Loss of Data:** Users lose their adjusted crop settings when navigating away or refreshing the page.
- **Inconsistent Delivery:** The public-facing website cannot render the cropped version of the image because it lacks access to the specific coordinates selected in the admin panel.
- **Manual Repetition:** Content managers must re-crop images every time they update a post if they want to maintain a specific visual focus.

### 1. Database Schema Update
- Add a `crop_metadata` field.
- The object should include: `x` (number), `y` (number), `width` (number), and `height` (number).

### 2. API Enhancement
- **Update Endpoint:** Modify the `POST/PATCH` endpoints to accept `crop` data within the payload.
- **Fetch Endpoint:** Ensure the `GET` endpoints return the `crop` metadata alongside the image URL.
- **Validation:** Implement server-side validation to ensure coordinates are within the actual dimensions of the uploaded image.

### 3. Frontend Integration
- Update the `onApply` handler in `PhotoBlock.tsx` to send the `MediaModalResult.crop` data to the backend during the save process.
- Ensure that when a record is loaded, the `savedCrop` state is initialized with data from the API response instead of `null`.

## Acceptance Criteria
- **Persistence:** Crop coordinates are successfully saved to the database and survive a page refresh.
- **State Restoration:** Re-opening a previously cropped image in the `MediaModal` correctly initializes the crop box using coordinates fetched from the API.
- **API Reliability:** The backend correctly handles cases where `crop` data is `null` (e.g., for uncropped images).
- **Data Integrity:** Saved coordinates accurately represent the `naturalWidth` and `naturalHeight` ratios to ensure cross-device consistency.

## Linked Issues
- Relates to: #405 (Resolve image cropping interaction issues)
- Follow-up to: #424 (Deprecate CropperModal)

Issue #424, opened 2026-03-31
(SP: 1) [Refactor] Deprecate CropperModal and unify image cropping within MediaModal


Fedorieieva:
## Description
During the resolution of issue **#405**, it was identified that the project contains a redundant `CropperModal` component. To improve maintainability and ensure a single source of truth for media interactions, we need to completely remove `CropperModal` and migrate all its remaining triggers to the enhanced `MediaModal`.

## Problem Statement
Having two different modals for cropping leads to:
1. **Code Duplication:** Maintenance of two separate sets of cropping logic and UI styles.
2. **Inconsistent UX:** Different interaction patterns for the same action (cropping) across the admin panel.
3. **State Management Overhead:** Potential conflicts in how crop coordinates are handled and saved.

## Steps
1. **Audit Dependencies:**
   - Identify all instances where `CropperModal` is imported and used across `lf-admin`.
2. **Migration:**
   - Replace `CropperModal` calls with `MediaModal` in the News/Events editor and other sections.
   - Ensure `MediaModal` receives the necessary props (`aspectRatio`, `initialCrop`, `onSave`) to fulfill the requirements previously handled by `CropperModal`.
3. **Decommissioning:**
   - Delete the `CropperModal` component folder.
   - Remove any associated styles or utility files used exclusively by this modal.

## Acceptance criteria
- [ ] `CropperModal.tsx` and its styles are completely removed from the repository.
- [ ] All features previously using `CropperModal` now trigger `MediaModal` without loss of functionality.
- [ ] Image cropping remains functional and accurate as per the "Definition of Done" in #405.
- [ ] No broken imports or dead code related to `CropperModal` remain in the project.

Issue #405, opened 2026-03-24
(SP: ) Resolve image cropping application and interaction issues


Jevgan:
## Description 
The image cropping functionality is currently experiencing several state management and interaction bugs across the `MediaModal` view. The core issues revolve around the crop boundaries not accurately reflecting user input, failing to reset, and lacking basic resize capabilities. 

## Steps to Reproduce 
     
1. **Broken Reset/Iteration Action:**
    - Clicking the "Iteration" icon inside the `MediaModal` is supposed to reset the crop to its default state, but it currently does nothing to the crop boundaries.
        
2. **Missing Resize/Scale Controls:**
    - Users are entirely unable to resize or scale the crop boundaries in the `MediaModal`. The crop box size is strictly locked, preventing users from cropping smaller or larger specific areas of the image.

## Definition of done

- **Accurate Cropping:** Moving the crop box must result in the final cropped image matching the exact selected area.
- ** State Retention:** Re-opening the modal should accurately reflect the user's last saved boundary positions.
- **Functional Reset:** The Iteration icon in `MediaModal` must successfully clear/reset the crop boundaries back to their default dimensions and position.
- **Draggable/Scalable Boundaries:** The user must be able to click and drag the corners/edges of the crop box to dynamically resize and scale the cropping area in both oval and rectangular modes.

PR #444, opened 2026-04-08
feat(api): implement persistent storage for image crop coordinates

Fedorieieva:
## Description

This PR implements the backend infrastructure and frontend integration required to persist image crop coordinates (`x`, `y`, `width`, `height`) across the application. Previously, crop state was volatile and lost upon page refresh. This solution ensures that user-defined crop boundaries are stored in the database and correctly restored during editing sessions.

**Key changes:**

* **Database & Schema Updates (Backend):**
    * **Common Schemas**: Added `CropRect` and `CropResult` types to the global schema. Updated `localizedImageSchema` to include a `crop` field, while deprecating legacy `width` and `height` fields that were stored outside the crop object.
    * **Base Content**: Extended the `BaseContent` interface and `LocalizedImage` type to include optional crop metadata, ensuring consistency across **News**, **Events**, and **Media Mentions**.
    * **Mongoose Models**: Integrated the new schema fields into `Page`, `News`, `Event`, and `MediaMention` models.
* **API Enhancements & Synchronization:**
    * **GraphQL API**: Updated GraphQL types and mutations (`UpsertPageDraft`, `PublishPage`, `Create/Update News/Events/MediaMentions`) to support the new `crop`.
    * **Sync Helpers**: Developed universal synchronization utilities in `helpers.ts`: `syncCoverImageCrop` (for main cover) and `syncContentImagesCrops` (recursively scans localized JSON content blocks).
    * **Data Persistence**: Implemented `syncImageCrops` logic within resolvers to handle data persistence in the dedicated `image_crops` collection.
* **Frontend Integration:**
    * **State Management**: Refactored `handleUploadImage` in `uploadToTmpFolder.ts` to perform a shallow copy of existing state, preventing the "state-overwriting" bug where `alt`, `caption`, and `crop` were lost.
    * **PhotoBlock Integration**: Updated the `onApply` handler in `PhotoBlock.tsx` to pass `MediaModalResult.crop` to the backend.
    * **State Restoration**: Ensured that the `savedCrop` state is initialized with coordinates fetched from the API for perfect restoration in `MediaModal`.
* **Testing & Stability:**
    * **Unit Tests**: Enhanced `uploadToTmpFolder.test.ts` for metadata validation and updated resolver tests with a new `createMockContext` utility.
    * **ESM Support**: Fixed Mongoose initialization and mock structures in tests to prevent `Unexpected token 'export'` errors related to ESM/BSON parsing.

**Technical Notes:**
The implementation uses a dual-storage approach: Coordinates are stored **inline** within content blocks for fast rendering and **centralized** in the `image_crops` collection for global asset management.

> [!IMPORTANT]
> **Known SonarCloud Warning:** The warning on `src/__mocks__/mongoose.ts` regarding unnecessary assertion is a **false positive**. Removing it causes `TS2339` errors and breaks the test suite. The assertion is required for the TypeScript compiler.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring 
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. **Persistence Test:**
    * Upload an image and set specific crop boundaries.
    * Save the changes and refresh the page.
    * Verify that crop coordinates survive and are visible in the database.
2. **State Restoration Test:**
    * Re-open the `MediaModal` for a previously cropped image.
    * Verify that the crop box is correctly initialized using the saved API data.
3. **API Reliability:**
    * Save an image without any cropping.
    * Ensure the backend handles `null` crop data without errors.
4. **Data Integrity:**
    * Check if saved coordinates accurately represent the `naturalWidth` and `naturalHeight` ratios.

## Video / Screenshots

**After**

https://github.com/user-attachments/assets/10b0a5eb-7cd9-413d-beb8-7dc3da5c6206

Closes [#428](#428)
Relates to [#405](#405)
Follow-up to [#424](#424)


Jevgan (approved):

src/interfaces/graphql/resolvers/helpers.ts
```diff
@@ -89,4 +92,71 @@ export const processSlugUpdate = async <
       }
     });
   }
+};
+
+export const syncCoverImageCrop = async (
+  contentId: string,
+  image: LocalizedImage,
+  locale: 'uk' | 'en' = 'uk'
```
kandyba: `locale` defaults to '`uk`' and none of the callers ever pass '`en`'. This means the crop for the English version of coverImage is never persisted to `image_crops`.

Is this intentional (because the cover is the same for both locales), or should we sync both?

kandyba (commented):
Three functions write to the `image_crops` collection, each using a different filter key strategy:

- `syncImageCrops` (PageMutation.ts): `{ cropId: item.src, pageId } `- cropId = filename (e.g. "`photo.jpg`")

- `syncCoverImageCrop` (helpers.ts): `{ pageId, cropId: 'coverImage', locale }` - cropId = fixed string

- `syncContentImagesCrops` (helpers.ts): `{ pageId, blockId, locale } `- uses `blockId` field instead of `cropId`

This can cause two issues:
1. Duplicates- the same image crop may be saved twice with different keys (e.g. function 1 saves by filename, function 2 saves by '`coverImage`'), resulting in two records for the same crop

2. Orphaned lookups - a crop saved by one function can't be found/updated by another, since they query by different fields. For example, `syncImageCrops` saves with `{ cropId: "photo.jpg" }`, but `syncContentImagesCrops` would look for `{ blockId: "content.intro.image" } `and create a new record instead of updating the existing one

Could we align on a single key schema, e.g. `{ pageId, cropId: "<block.path.to.image>", locale }`, so all three functions can reliably find and update the same record?

As example: `{ pageId: "abc123", cropId: "IntroSection.image", locale: "uk" }`

src/interfaces/graphql/resolvers/helpers.ts
Fedorieieva: Thank you for pointing this out, you are right. Currently, the image_crops collection is not yet actively used for rendering or lookups, so this wasn't causing immediate issues. However, for the sake of architectural consistency and to ensure the MediaModal initializes correctly regardless of the selected locale in the future, I have taken your feedback into account. I've updated the helper to sync the crop data for both locales simultaneously. Fixed in the latest commi

DenisGordProgrammer (changes requested):
Nice PR! However Add Video file like an file, not as link, please

DenisGordProgrammer (approved):

src/application/use-cases/extractImageSrc/extractImageSrc.ts
```diff
@@ -1,21 +1,41 @@
...
     }
-    if (typeof value === 'object') {
-      const obj = value as { [key: string]: unknown };
-      if (typeof obj.src === 'string' && obj.isTmp === true) {
-        sources.add(obj.src);
+
+    const obj = value as Record<string, unknown>;
+
+    if (typeof obj.src === 'string') {
+      if (obj.isTmp === true || obj.crop) {
```
kandyba: `extractImageSrcs` now returns images that have `crop` but not` isTmp: true.` However, in `PageMutation.ts` (and potentially other places) the result is passed directly to `copyBlobsToNewFolder('tmp', 'photos', imageSrcs)`, which assumes all returned files exist in the `tmp` folder

If an admin only re-crops an existing published image (no re-upload), the file won't be in `tmp` this will either fail or silently skip the copy

Before we had condition:
```
if (typeof obj.src === 'string' && obj.isTmp === true) {
  sources.add(obj.src);
}
```

Should `extractImageSrcs` stay focused on tmp-only images for the copy use case, while `extractImagesWithMetadata` handles the crop sync separately?

src/interfaces/graphql/resolvers/helpers.ts
```diff
@@ -89,4 +94,66 @@ export const processSlugUpdate = async <
...
+  data: unknown,
+  options: {
+    locale?: 'uk' | 'en';
+    cropIdPrefix?: string;
+    isCoverImage?: boolean;
+  } = {}
+): Promise<void> => {
+  const { locale = 'uk', cropIdPrefix = '', isCoverImage = false } = options;
+
+  if (!isCoverImage) {
```
kandyba, line 109: For the cover image mode, both locales are now synced- nice fix. But in content mode, even though `locale` can technically be passed via `options`, none of the callers actually provide it (`eventsMutation, newsMutation, PageMutation` all call `syncImagesCrops(id, data)` without the third argument), so it always defaults to '`uk`'. Is this intentional because content images share the same crop across locales, or should we either pass both locales explicitly or apply the same both-locales approach as in cover mode?

Maybe we can set `local` as a third parameter?

`await syncImagesCrops(res.id, input.content, { locale: 'en' });`

src/interfaces/graphql/resolvers/helpers.ts
```diff
@@ -89,4 +94,66 @@ export const processSlugUpdate = async <
...
+      await Promise.all(locales.map(l =>
+        ImageCropModel.findOneAndUpdate(
+          {
+            pageId: contentId,
+            cropId: cropIdPrefix || 'coverImage',
+            locale: l
+          },
+          {
+            crop: image.crop,
+            imageAssetId: new Types.ObjectId(),
```
kandyba: `imageAssetId` is set to new `Types.ObjectId() `on every upsert, so it gets a new random value each time the crop is saved. This means:

- It doesn't reference any actual document in the `assets` collection (or any other collection)- it's generated on the spot
- It's overwritten on every save if an admin saves crop twice, the previous `imageAssetId` is lost
- No code in this PR reads or queries by `imageAssetId`, so it seems unused

Is this a placeholder for a future feature (e.g. linking crops to assets)? If so, a TODO comment would help. If the schema requires it, should we pass the real asset ID instead? And if it's not needed, removing it would reduce confusion

kandyba (changes requested):
Nice cleanup unifying the sync functions. Three remaining questions:

1. `extractImageSrcs` + blob copy (critical): Now returns images with `crop` but without `isTmp`. PageMutation passes these to `copyBlobsToNewFolder('tmp', ...)` — files that were only re-cropped (not re-uploaded) won't exist in `tmp`.

2. Content mode locale: Cover mode now syncs both locales, but all content-mode callers omit `locale`, so it always defaults to '`uk`'. Intentional?

3. `imageAssetId`: Random `ObjectId` on every upsert, not linked to any document, never queried. Placeholder for future use or can be removed?

src/interfaces/graphql/resolvers/helpers.ts
Fedorieieva: Generating a new Types.ObjectId() on every upsert was indeed a 'time bomb' for unique indexes and created orphan IDs.

I have implemented the following changes to address this:
1. Removed imageAssetId from the syncImagesCrops logic in helpers.ts.
2. Updated imageCrop.model.ts by removing the required constraint and the ref: 'ImageAsset' from imageAssetId.
3. ICleaned up the incorrect indexes related to imageAssetId and added a proper unique index on the natural key: { pageId, cropId, locale }.

Regarding the transition to a polymorphic reference pattern (e.g., contentId + contentType): while I agree this would be a superior architectural fit, it involves significant structural changes and requires a dedicated data migration. To maintain the stability of the current PR, I’ve focused on fixing the immediate indexing issues; a full architectural refactor can be handled in a separate, dedicated pull request if necessary

src/interfaces/graphql/resolvers/helpers.ts
Fedorieieva: Regarding the locale in content mode: this is intentional. Unlike cover images, which might have specific localized requirements, content images and their crops are universal across locales in our current architecture. Rendering and cropping logic for these images does not depend on the language context. Therefore, defaulting to 'uk' serves as a stable fallback for the storage key without needing redundant synchronization for multiple locales

src/application/use-cases/extractImageSrc/extractImageSrc.ts
Fedorieieva: You're right, this was a regression. I have reverted extractImageSrcs to its original behavior: it now strictly returns only images where isTmp === true. This ensures that PageMutation only attempts to copy files that actually exist in the temporary storage. extractImagesWithMetadata remains separate and handles all images that require crop synchronization, regardless of their temporary status

src/interfaces/graphql/resolvers/helpers.ts
kandyba: @Fedorieieva 
Makes sense

Just to confirm when we later read these crops back, the consumer will always query with locale: 'uk' for content images? If so, might be worth adding a brief comment in the code explaining this convention, so future devs don't accidentally query with 'en' and get empty results.
What do you think?

kandyba (approved):
Great work on this PR- persistent storage for crop coordinates with the dual-storage approach is a solid design. Thank you for thoroughly addressing all the review feedback.

One small note: I left a comment about the hardcoded locale: 'uk' for content mode- since it's intentional, a brief code comment explaining this convention would help avoid confusion for future contributors

PR approved ✅

PR #456, opened 2026-04-13
fix(ui): apply crop coordinates to image preview in PhotoBlock

### LINKED ISSUE #455: (SP: 2) [Bug] Images on cards ignore crop coordinates
## Summary

Images inside the page block cards components are currently rendered in their full original scale or stretched to fit the container. The **crop coordinates** (x, y, width, height) stored in the database are being ignored, resulting in incorrect visual framing of the content on the public website.

Image

## Expected behavior

The card should use the stored metadata to visually crop the image. The user should see the exact area they selected in the `MediaModal` during the editing process.

Fedorieieva:
## Description

This PR fixes an issue where images in the `PhotoBlock` component ignored the crop metadata stored in the database. I implemented the `useCroppedImage` hook to calculate the correct CSS transformations (scale and translate) based on the saved crop coordinates (x, y, width, height). This ensures that the preview in the admin panel matches the framing selected by the user in the MediaModal.

Summary of changes:
- Integrated `useCroppedImage` hook into `ImagePreviewBlock`.
- Updated `PhotoBlock.tsx` to apply calculated styles to the image preview.
- Ensured images are no longer stretched or displayed in full scale when a crop is present.

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: ## How to test

1. Open the admin panel and navigate to a section containing a `PhotoBlock` (about-us page).
2. Upload an image or select an existing one.
3. Open the "Edit" (Редагувати) mode and set a specific crop area.
4. Save the changes and verify that the preview image in the `PhotoBlock` accurately displays only the cropped area.

### Before

https://github.com/user-attachments/assets/b6f34cfa-5137-45a5-84a2-5e54e5984450

### After

https://github.com/user-attachments/assets/fe6f0ffa-3d91-475d-80f5-4f2dadc7a442

Closes #455

### DIFF
--- a/app/shared/components/design-system/photo-block/PhotoBlock.tsx
+++ b/app/shared/components/design-system/photo-block/PhotoBlock.tsx
@@ -6,12 +6,16 @@ import toast from 'react-hot-toast';
 
 import Button from '../button/Button';
 import { styles } from './PhotoBlock.styles';
+import { useCroppedImage } from '~/hooks/use-cropped-image/use-cropped-image';
 import { readFileAsDataURL } from '~/lib/utils/readFileAsDataURL';
 import ImageIcon from '~/public/icons/image.svg';
 import PencilIcon from '~/public/icons/pencil.svg';
 import { MediaModal } from '~/shared/components/media-modal/MediaModal';
 import type { MediaModalOpenState, MediaModalResult } from '~/shared/components/media-modal/MediaModal.types';
 import { useImageMetadata } from '~/shared/hooks/use-image-metadata/useImageMetadata';
+
+const PREVIEW_W = 196;
+const PREVIEW_H = 120;
 
 interface ImagePreviewBlockProps extends StackProps {
   imageUrl: string;
@@ -48,18 +52,23 @@ export const ImagePreviewBlock = ({
 
   const [isMediaModalOpen, setIsMediaModalOpen] = useState(false);
   const [mediaInitial, setMediaInitial] = useState<MediaModalOpenState | undefined>(undefined);
-
-  const [savedCrop, setSavedCrop] = useState<MediaModalResult['crop']>(initialCrop || null);
+  const [savedCrop, setSavedCrop] = useState<MediaModalResult['crop']>(initialCrop ?? null);
 
   useEffect(() => {
     setPreviewImage(imageUrl);
   }, [imageUrl]);
 
   useEffect(() => {
-    setSavedCrop(initialCrop || null);
+    setSavedCrop(initialCrop ?? null);
   }, [initialCrop]);
 
   const { dimensions, fileName: finalFileName } = useImageMetadata(previewImage, fileName);
+
+  const { styles: cropStyles, onLoad: onImgLoad } = useCroppedImage(
+    savedCrop,
+    PREVIEW_W,
+    PREVIEW_H
+  );
 
   const openEditCrop = () => {
     setMediaInitial({
@@ -125,12 +134,24 @@ export const ImagePreviewBlock = ({
 
       <Box sx={styles.imageBlock}>
         {previewImage ? (
-          <Box
-            component="img"
-            src={previewImage}
-            alt={title || 'Selected'}
-            sx={oval ? styles.imageOvalPreview : styles.imagePreview}
-          />
+          oval ? (
+            <Box
+              component="img"
+              src={previewImage}
+              alt={title || 'Selected'}
+              sx={styles.imageOvalPreview}
+            />
+          ) : (
+            <Box sx={cropStyles.container}>
+              <Box
+                component="img"
+                src={previewImage}
+                alt={title || 'Selected'}
+                onLoad={onImgLoad}
+                sx={cropStyles.image}
+              />
+            </Box>
+          )
         ) : (
           <Box sx={styles.imagePreview}>
             <Box
--- a/app/shared/components/design-system/photo-block/PhotoBlock.tsx
+++ b/app/shared/components/design-system/photo-block/PhotoBlock.tsx
@@ -6,7 +6,7 @@ import { useEffect, useMemo, useState } from 'react';
 import toast from 'react-hot-toast';
 
 import Button from '../button/Button';
-import { styles } from './PhotoBlock.styles';
+import {PREVIEW_H, PREVIEW_W, styles} from './PhotoBlock.styles';
 import { useCroppedImage } from '~/hooks/use-cropped-image/use-cropped-image';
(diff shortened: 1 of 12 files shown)


app/shared/components/design-system/photo-block/PhotoBlock.tsx
```diff
@@ -115,6 +124,44 @@ export const ImagePreviewBlock = ({
     }
   };
 
+  const renderPreviewContent = useMemo(() => {
+    if (!previewImage) {
+      return (
+        <Box sx={styles.imagePreview}>
+          <Box
+            component="img"
+            src="/icons/cloud-upload.svg"
```
Jevgan: According to our convenience to use icons from lucide, please change usage of `Box[src="..."]` to lucide import:
```
import { CloudUpload } from 'lucide-react';
...
<CloudUpload />
```

DenisGordProgrammer (approved):

Jevgan (approved):

app/shared/hooks/use-cropped-image/use-cropped-image.ts
```diff
@@ -0,0 +1,87 @@
...
+  const containerBase: React.CSSProperties = {
+    width: containerW,
+    height: containerH,
+    overflow: 'hidden',
+    position: 'relative',
+    flexShrink: 0,
+    border: '1px solid #B2B3BE',
+  };
+
+  if (!crop?.rect || natW === 0 || natH === 0) {
```
kandyba: `crop.rect` can exist but have `width: 0` or `height: 0` (e.g. a click without drag, or corrupted data from the database). This guard doesn't cover that case, so lines 41-42
```
 const scaleX = containerW / cropW;
 const scaleY = containerH / cropH;
```
would divide by zero and produce Infinity in the CSS transform and the image just silently disappears

A small addition to the condition should have something like:
`crop.rect.width === 0 || crop.rect.height === 0`

What do you think?

app/shared/hooks/use-cropped-image/use-cropped-image.ts
```diff
@@ -0,0 +1,87 @@
+import { useCallback, useMemo, useState } from 'react';
+
+import type { CropResult } from '~/shared/components/media-modal/MediaModal.types';
```
kandyba: This hook is a shared utility, but it imports `CropResult` from a specific component `MediaModal.types`. The same type already exists in` ~/types/common` other parts of the codebase (e.g. `OurMission.tsx`) import it from there. Would be better to keep the same convention:
`import type { CropResult } from '~/types/common';`
This way the hook doesn't depend on a specific component.

But now we have duplicated `export interface CropRect`

app/shared/components/design-system/photo-block/PhotoBlock.tsx
```diff
@@ -1,18 +1,23 @@
...
+import { useCroppedImage } from '~/hooks/use-cropped-image/use-cropped-image';
 import { readFileAsDataURL } from '~/lib/utils/readFileAsDataURL';
 import ImageIcon from '~/public/icons/image.svg';
 import PencilIcon from '~/public/icons/pencil.svg';
 import { MediaModal } from '~/shared/components/media-modal/MediaModal';
 import type { MediaModalOpenState, MediaModalResult } from '~/shared/components/media-modal/MediaModal.types';
 import { useImageMetadata } from '~/shared/hooks/use-image-metadata/useImageMetadata';
 
+const PREVIEW_W = 196;
+const PREVIEW_H = 120;
```
kandyba: These values duplicate what's already in `PhotoBlock.styles.ts` file `imagePreview.width: '196px', imagePreview.height: '120px'`
If someone changes the size in one place and forgets the other, the crop preview will break silently.

Maybe it better to update styles using const-s?
```
imagePreview: {
  width: PREVIEW_W,
  height: PREVIEW_H,
}
```

app/shared/hooks/use-cropped-image/use-cropped-image.ts
```diff
@@ -0,0 +1,87 @@
...
+  containerW: number,
+  containerH: number
+): CroppedImageStyles {
+  const containerBase: React.CSSProperties = {
+    width: containerW,
+    height: containerH,
+    overflow: 'hidden',
+    position: 'relative',
+    flexShrink: 0,
+    border: '1px solid #B2B3BE',
```
kandyba: `overflow: 'hidden'` and `position: 'relative' `make sense here- they're needed for the crop to work. But `border` is purely visual and not related to cropping logic. It also duplicates the border from `PhotoBlock.styles.ts`

I'd remove `border` from the hook and let the consuming component handle it. This way the hook stays focused on crop math and can be reused in other places with different styling

app/shared/hooks/use-cropped-image/use-cropped-image.ts
```diff
@@ -0,0 +1,87 @@
...
+      height: natH,
+      maxWidth: 'none',
+      transformOrigin: '0 0',
+      transform: `translate(${translateX}px, ${translateY}px) scale(${scale})`,
+      display: 'block',
+    },
+  };
+}
+
+export function useCroppedImage(
```
kandyba, line 74: All other hooks in `hooks` have tests- could you add a test file for this one too? The math in `buildCroppedStyles` is a good candidate for unit testing since it's pure input/output with no DOM dependencies

What do you think?

app/shared/hooks/use-cropped-image/use-cropped-image.ts
```diff
@@ -0,0 +1,87 @@
...
+    setNaturalSize({ w: img.naturalWidth, h: img.naturalHeight });
+  }, []);
+
+  const styles = useMemo(
+    () => buildCroppedStyles(crop, naturalSize.w, naturalSize.h, containerW, containerH),
+    [crop, naturalSize.w, naturalSize.h, containerW, containerH]
+  );
+
+  return { styles, onLoad };
+}
```
kandyba: Nit: missing newline at end of file

kandyba (changes requested):
Great job on this one! 

This code is a solid addition. I left a few comments inline please take a look when you get a chance

app/shared/hooks/use-cropped-image/use-cropped-image.ts
Fedorieieva: Thanks! That's a game changer😊

kandyba (approved):
Looks awesome! Thanks

PR #495, opened 2026-04-30
feat(upload): allowed to upload different file types

### LINKED ISSUE #491: (SP: )  Allow uploading multiple file types (PDF, Audio, Docs) in Files page
## Description

Currently, the system only supports image uploads. The task is to extend both frontend and backend validation to support additional document, audio, and archive formats (PDF, DOCX, XLSX, MP3, WAV, ZIP, RAR).

## Acceptance Criteria
- Update FILES_UPLOAD_ALLOWED_MIME_TYPES and FILES_UPLOAD_ALLOWED_EXTENSIONS in the frontend constants.
- Modify the backend validator for the /api/uploads/single endpoint to include new MIME types (including handling OS-specific .rar MIME types).
- Update the MongoDB Asset schema enums to accept the new file types.
- Ensure MediaModal and UploadView correctly handle and display the selected non-image files (e.g., bypassing the image cropper for documents/archives).
- Successful upload returns a valid Cloudflare R2 URL for the new file types.

## Expected Result

1. The file passes client-side validation in the UploadView component.
2. The backend API /api/uploads/single accepts the file and returns a 201 Created (or 200 OK) status.
3. The uploaded file appears in the file list with a valid Cloudflare R2 Storage URL.
4. The file is fully accessible and downloadable via the sidebar link.

dest411:
## Description

Implemented backend infrastructure and API support for multi-type file uploads (PDF, Documents, Archives, Audio, Video). This PR synchronizes validation and storage logic across all layers.

## Type of change

- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Open the admin panel on the files page so the client fetches files from the storage.

2. Open DevTools, go to the Network tab, and find the request named graphql to locate the Cookie containing the accessToken. Copy this accessToken in the format accessToken=eyJhbG... so you can pass it into Postman to authorize your request (see screenshot).

3. Open Postman, create a new tab, select the POST method, and enter http://localhost:3000/api/uploads/single as the endpoint URL. Go to the Headers tab, add Key: Cookie, and paste the copied accessToken into the Value field.

4. Go to the Body tab (select form-data) and add:
      - Key: directory, Value: works
      - Key: fileType, Value: spreadsheet (Note: this is for the xlsx format. For other formats, use document, archive, or audio   accordingly).
      - Key: validationRules, Value: {"allowedMimeTypes":["application/vnd.openxmlformats-officedocument.spreadsheetml.sheet"],"allowedExtensions":["xlsx"]} (This is the validation rule required by Postman for xlsx. You need to specify appropriate values for other formats).
      - Key: file — change the type from text to file, and in the Value field, upload a file from your computer.

5. Click Send and verify you receive a 201 Created status along with a JSON response.
image

6. Create a new request tab in Postman, select the POST method, and enter http://localhost:3000/api/graphql as the endpoint URL. Go to the Headers tab, add Key: Cookie, and paste the same accessToken from the previous step. In the same tab, add Key: apollo-require-preflight, Value: true.

7. Go to the Body tab and select the GraphQL option. In the Query field, paste the following mutation:

`GraphQL: `
```
mutation CreateAsset($input: CreateAssetInput!) {
  createAsset(input: $input) {
    id
    filename
    url
    mimeType
    type
  }
}
```
In the GraphQL Variables field, paste a JSON payload formed from the data received in the previous request. Example:

`JSON:`
```
{
  "input": {
    "filename": "1777645231273-cfc3911139c3bd15.xlsx",
    "mimeType": "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
    "sizeBytes": 6635,
    "type": "spreadsheet", 
    "url": "https://pub-2b50c59c64954ab89b7837f9f4607e12.r2.dev/works/1777645231273-cfc3911139c3bd15.xlsx"
  }
}
```
(Note: you can use AI to quickly generate this JSON payload based on the previous response).

8. Click Send and verify you receive a 200 OK status with a valid JSON response. Finally, check MongoDB to ensure the file record was successfully created.

## Expected result
- The system successfully validates, uploads, and registers PDF, Documents (Docx/Xlsx), and Archives.
- The database remains consistent with the GraphQL schema.

Closes: https://github.com/Liatoshynsky-Foundation/lf-admin/issues/491


src/uploads/validators/validatorFactory.ts
```diff
@@ -1,41 +1,63 @@
...
+
+      const sizeValid = validateFileSize(buffer, rules.maxSize);
+      const mimeValid = validateMimeType(mimeType, rules.allowedMimeTypes);
+      const extValid = validateExtension(filename, rules.allowedExtensions);
+
+      return combineValidationResults(sizeValid, mimeValid, extValid);
+    }
+  };
+};
+
```
Fedorieieva: I’ve reviewed the changes to validatorFactory.ts, and I'd like to discuss the approach with the createGenericValidator.

While I understand the intention to support multiple new file types, I believe it's better to implement specific validators for different categories rather than falling back to a generic validator for everything. Here’s why:

1. different file types often require specific validation rules. For example, image validation checks dimensions, and other types might require their own specific checks in the future.

2. the original validatorFactory.ts distinguishes between file types and provides specific errors (e.g., DOCUMENT_VALIDATOR_NOT_IMPLEMENTED). Using a generic fallback hides when a specific validator has not been implemented yet.

3. keeping validation logic modular allows us to unit-test each file type independently without overloading a single function.

src/uploads/validators/validatorFactory.ts
```diff
@@ -1,41 +1,63 @@
...
+  case 'archive':
   case 'generic':
-    // OPTIONAL: Implement generic file validator
-    // return createGenericValidator(config.rules);
-    throw new Error(UPLOAD_ERRORS.GENERIC_VALIDATOR_NOT_IMPLEMENTED);
+    return createGenericValidator(config.rules);
 
   default:
-    throw new Error(`${UPLOAD_ERRORS.UNKNOWN_FILE_TYPE}: ${config.fileType}`);
+    return createGenericValidator(config.rules);
```
Fedorieieva: In addition I strongly recommend keeping the error for unknown file types so the system fails immediately rather than silently falling back to generic validation. This approach improves debugging visibility for integration issues and ensures strict security and data integrity by explicitly controlling the supported formats.

src/uploads/initialize.ts
```diff
@@ -14,8 +13,8 @@ export const initializeUploadModule = (config: Config) => {
 
   const uploadService = createUploadService({
     storage,
-    defaultFileType: 'image',
-    defaultValidationRules: DEFAULT_IMAGE_RULES as Record<string, unknown>
+    defaultFileType: 'generic',
+    defaultValidationRules: {}
```
Fedorieieva: I suggest reverting the changes made to initialize.ts. Here are the main reasons why we shouldn't use 'generic' and an empty object {} as defaults:

1. setting defaultValidationRules to an empty object removes all file constraints (size, mime type, etc.). This allows users to upload potentially malicious or oversized files without restrictions.

2. the upload service relies on defaultFileType when no options are provided. It should default to a secure, well-defined type (like 'image', which already has validation rules and logic implemented).

3. a generic type without rules bypasses the validation mechanism designed to protect our storage and application.

app/shared/components/media-modal/FileUploadModal.tsx
Fedorieieva, line 1: I'd like to suggest a few architectural improvements to make this feature more modular and consistent with our existing codebase.

Here are my thoughts on how we can refine the approach:

1. instead of creating the component directly within the media-modal directory, it would be better to add a dedicated view (e.g., file-view in views directory). This keeps the module isolated and aligns with our existing pattern for views.

2. to ensure the component is reusable across different layouts, we could introduce a prop (for example, to control the header visibility or layout variant) rather than hardcoding the logic inside the component.

Fedorieieva (changes requested):
Thanks for adding new code to the file upload functionality! I highly recommend splitting this PR into two separate tasks: one for the backend validation logic and one for the frontend UI components. Separating these layers allows for a more focused review, ensures a clean separation of concerns, and makes it much safer and easier to test and deploy changes without cross-domain issues

src/uploads/validators/validatorFactory.ts
dest411: Makes sense. I've removed the generic fallback and implemented specific validators (for documents, archives, audio, etc.) to ensure modularity.

src/uploads/validators/validatorFactory.ts
dest411: Done. Restored the strict throw new Error in the default case

src/uploads/initialize.ts
dest411: Totally agree, empty rules were a security risk. Reverted to the original secure defaults ('image' and strict rules)

app/shared/components/media-modal/FileUploadModal.tsx
dest411: Great suggestion. I will implement this dedicated FileView approach with layout props in the upcoming frontend PR.

Th0mas-H0ward (changes requested):
Tested it locally. The upload works correctly for most file types. However, the PR description does not mention support for file types such as XLSX, MP3, and WAV, even though this is required in the ticket. Nevertheless, uploading MP3 and WAV files works correctly. However, uploading XLSX files results in a backend validation error (see screenshots). Therefore, please correct the PR text and investigate the error with XLSX files.
Anyway, you did a great job! Waiting for your updates to check it again 💪

Image


Image


image

Th0mas-H0ward (approved):
I've checked it again, and everything works as intended - XLSX files can be uploaded without issues. Please consider adding the additional media formats as we discussed, update PR text and pay attention to Daria's comments regarding the code.

DenisGordProgrammer (approved):

Fedorieieva (approved):

PR #579, opened 2026-06-03
feat(logs): add configurable retention and manual cleanup

### LINKED ISSUE #540: (SP: 2) [Feature][Logging] Add logs auto-cleanup and retention strategy
### Summary
Add automatic cleanup for application logs stored in MongoDB to prevent unlimited growth of the `logger` collection.

---

#### Problem
Centralized application logging has already been implemented in the previous task:

https://github.com/Liatoshynsky-Foundation/lf-admin/issues/518

Logs are stored in MongoDB collection `logger`, but there is currently no cleanup or retention strategy.

Over time, the logs collection may grow too much and affect database size, performance, and maintenance.

#### Goal
Introduce a simple log retention mechanism for stored application logs.

Possible approaches:
- auto-delete logs older than a configured period
- limit maximum number of stored log records
- optionally add manual `Clear logs` action in the admin logs page

---

### Scope

- Add cleanup / retention logic for MongoDB logs
- Prevent unlimited growth of the `logger` collection
- Make retention configurable where possible
- Optionally add manual clear action for admins
- Update README with short retention/cleanup info
- Keep existing `/logs` page functionality working

---

### Acceptance Criteria

- [ ] Logs cleanup / retention strategy is implemented
- [ ] `logger` collection does not grow without limits
- [ ] Existing `/logs` page continues to work correctly
- [ ] README is updated with retention/cleanup information

---

### QA Checklist

- [ ] Generate several logs
- [ ] Verify logs are saved in MongoDB
- [ ] Verify cleanup removes logs according to the selected strategy
- [ ] Verify `/logs` page still displays available logs correctly
- [ ] Verify filters and pagination still work

---

### Risks / Known Limitations

- Cleanup strategy should avoid deleting logs too aggressively
- Manual clear action, if added, should be protected from accidental use
- Retention period / max records may need adjustment later based on real log volume

---

### Related

- https://github.com/Liatoshynsky-Foundation/lf-admin/issues/518

---

### Branch name

```bash
feat/XXX/add-logs-retention-cleanup

### PR DESCRIPTION
**Pull request by dmitryzh100**
## Description

Adds a retention/cleanup strategy for application logs so the MongoDB `logger` collection no longer grows  without limits.
  - **Configurable TTL** via `LOG_RETENTION_SECONDS` env var (defaults to `604800` — 7 days). The
  `winston-mongodb` transport creates a TTL index on `timestamp` so MongoDB auto-expires old logs in the
  background.
  - **Manual clear action** on the `/logs` admin page: a single black "Видалити …" button whose label and scope
  adapt to the active tab ("Видалити всі логи" on the *Усі* tab, "Видалити Error логи" / "Видалити Warn логи" /
  etc. on a level tab). Confirmation dialog, disabled when the visible list is empty.
  - **`DELETE /api/logs`** endpoint backing the action; optional `?level=` filter; gated by access-token cookie
  verification using the same `createTokenService` pattern as the GraphQL context.
  - **README** updated with the new env var, the auto-expiry behavior, and a note that an existing TTL index isn't  auto-updated on retention changes (collection/index must be dropped or `collMod`'d).

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Setup — start the app against any MongoDB instance you have access to. Defaults give 7-day retention; set
  `LOG_RETENTION_SECONDS` in your env to override. 
2. Manual clear (Phase 1) — log in, use the app to generate some logs, open `/logs` page.  Verify:
    - On the "Усі" tab the button reads "Видалити всі логи" and clears the entire collection.
    - On a level tab (Error / Warn / Info / Debug) the button reads "Видалити  логи" and clears only that level.
    - The button is disabled when the visible list is empty.
    - The action requires a valid admin session (DELETE returns 401 otherwise).
  3. TTL retention (Phase 2) — set LOG_RETENTION_SECONDS=60, drop the existing logger collection so MongoDB
  recreates the TTL index with the new value, restart the app, write a few logs. Within ~60–120 s the collection
  empties on its own and /logs shows no records.
  4. Reset — restore LOG_RETENTION_SECONDS to your normal value before continuing other work.

## Video / Screenshots

Manual logs deletetion from admin panel
https://github.com/user-attachments/assets/45debc49-2981-4e98-ad92-c6e887069b72

Automatic logs retention (retention period set to 60 seconds)
https://github.com/user-attachments/assets/e07489d9-0798-4f1a-a445-bebf62ce710c

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues
- [x] Documentation updated (if applicable)

---
Issue: [540](https://github.com/Liatoshynsky-Foundation/lf-admin/issues/540)

### DIFF
--- a/app/(logged_in)/logs/LogsPageClient.tsx
+++ b/app/(logged_in)/logs/LogsPageClient.tsx
@@ -207,6 +207,8 @@ const LogsPageClient = () => {
 
       setPage(1);
       setReloadKey((value) => value + 1);
+    } catch (error) {
+      console.warn(error);
     } finally {
       setDialogOpen(false);
       setClearing(false);
--- a/app/(logged_in)/logs/LogsPageClient.tsx
+++ b/app/(logged_in)/logs/LogsPageClient.tsx
@@ -6,14 +6,20 @@ import {
   AccordionDetails,
   AccordionSummary,
   Box,
+  Button,
   Chip,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogContentText,
+  DialogTitle,
   Pagination,
   Stack,
   Tab,
   Tabs,
   Typography
 } from '@mui/material';
-import { type SyntheticEvent, useEffect, useState } from 'react';
+import { type SyntheticEvent, useCallback, useEffect, useState } from 'react';
 
 import type { LogEntry, LogLevel, LogsResponse } from '~/back-shared/types/logs';
 
@@ -133,6 +139,9 @@ const LogsPageClient = () => {
   const [items, setItems] = useState<LogEntry[]>([]);
   const [total, setTotal] = useState(0);
   const [loading, setLoading] = useState(false);
+  const [reloadKey, setReloadKey] = useState(0);
+  const [dialogOpen, setDialogOpen] = useState(false);
+  const [clearing, setClearing] = useState(false);
 
   useEffect(() => {
     const controller = new AbortController();
@@ -162,7 +171,7 @@ const LogsPageClient = () => {
 
     loadLogs();
     return () => controller.abort();
-  }, [level, page]);
+  }, [level, page, reloadKey]);
 
   const totalPages = Math.max(1, Math.ceil(total / 20));
   const hasItems = items.length > 0;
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

**Review by DenisGordProgrammer — APPROVED** [review]
Image

File: app/(logged_in)/logs/LogsPageClient.tsx [hunk 1]
```diff
@@ -173,6 +182,41 @@ const LogsPageClient = () => {
...
+        throw new Error('Failed to clear logs');
+      }
+
+      setPage(1);
+      setReloadKey((value) => value + 1);
+    } finally {
+      setDialogOpen(false);
+      setClearing(false);
+    }
+  }, [level]);
```
Yushchyk-Roman, line 216: Great job on this PR! The implementation is solid, and it fully covers the task requirements.

I noticed one minor issue that we should probably fix before merging, along with a couple of ideas for future improvements:

> [!WARNING]
> ### Good to Fix in this PR
> **Missing `catch` block in `confirmClear`:** Currently, there is only a `try...finally` structure here. If the `DELETE` request fails (e.g., due to a 401 or 500 error), it throws an error that bypasses `finally` and results in an `Unhandled Promise Rejection` in the browser console, while the modal just silently closes. Let's add a quick `catch (error) { console.warn(error); }` block to keep the error handled.

### Future Ideas (Not blocking this PR, for separate tickets)
1. **Auth UX (401 Error):** Inside the `catch` blocks, when we hit a `401 Unauthorized` (expired token), it would be a much better UX to automatically redirect the user to the login page instead of just silently clearing the state.
2. **Server Failures (500 Error):** When the server goes down, we can't send logs back to it, but we should definitely log it to the browser console (`console.error`) and maybe show a friendly Toast alert to the user.

Once the `catch` block is added, this is good to go!

Yushchyk-Roman (approved):

app/(logged_in)/logs/LogsPageClient.tsx
dmitryzh100: change have been added

PR #592, opened 2026-06-08
feat(ui): privacy-policy edit page 

### LINKED ISSUE #271: (SP: 2) Privacy Policy — Add Sections Editor in Admin
## Description

- Implement an editor for Privacy Policy page in the admin panel.

- Add support for creating, editing, and reordering sections of the Privacy Policy page.

- Each section should support localized fields (uk/en) and common block types (title, text, lists, links).

- Ensure the page supports draft/published states and can be previewed.

telare:
## Description

This PR implements the full editing functionality for the Privacy Policy page, allowing administrators to manage all
policy sections through the admin panel.

**Core implementation**
   * Added the main Privacy Policy editor page, which serves as the entry point for all policy-related modifications.
   * Implemented specific block components to cover every section of the policy
   * Defined the necessary TypeScript interfaces and block constants to support the new data structure in the store.

**Reusable Editor Components**
- EditParagraphsBlock — handles blocks made up of multiple formatted paragraphs.
- EditDescriptionListNoteBlock — covers a description, a dynamic list of points, and a closing note.
- PointsList — a focused UI component for managing dynamic lists of formatted text points.

**Custom Hooks**
- usePointsList — orchestrates adding, removing, and updating list items in the global store.
- useSectionList — handles the more complex nested structure (sections with subtitles and point lists) used in the "Data We Collect" section.

 **Testing**
   * Introduced edit-pages-factory.tsx to standardize tests for editor pages, including preview, save, and cancel actions.
   * Introduced block-test-factory.tsx to remove code duplication for repeated tests of individual editor blocks.

**IMPORTANT NOTE:** 
* In the code are no reordering features. 
* Also, you can see a lot of bad UX in the inputs for blocks. 
**I'll create dedicated tasks for these**, because the main purpose of this feature was to create this page. 

(Created related tasks: 
1. https://github.com/Liatoshynsky-Foundation/lf-admin/issues/603
2. https://github.com/Liatoshynsky-Foundation/lf-admin/issues/602)

Closes #271 

## Type of change

- [X] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Log in to the admin page
2. Open /privacy-policy 
3. Verify that the page is rendered successfully and has a similar behavior to the/about-us page
4. Verify that all blocks contain valid information about the page content for both EN and UK locales
5. Try to add, remove, or change list items or text fields and save the data by pressing the button at the top right corner. 
6. Verify that all necessary data has been changed and your changes have been applied at the admin page.
7. Go to the client privacy-policy page and verify your changes.
8. Run related tests using npm run test app/shared/components/privacy-policy & app/shared/hooks.

## Video / Screenshots

https://github.com/user-attachments/assets/70e22a5f-a655-42df-b22d-19312a0be12d

https://github.com/user-attachments/assets/c5d4a8c0-8cac-494d-abdc-12498142422f

## Checklist

- [X] Code compiles and runs correctly
- [X] Tests pass successfully
- [X] Linting checks are clean
- [X] No Sonar issues
- [X] Branch is up to date with the target branch
- [X] Linked issue/task included
- [X] Task moved to the review column on the board

---


app/(logged_in)/privacy-policy/page.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+      'data-retention',
+      'data-usage',
+      'data-we-collect',
+      'google-auth',
+      'intro-section',
+      'newsletter-subscription',
+      'social-networks',
+      'targeted-ads',
+      'user-rights',
+      'сookies'
```
DenisGordProgrammer: double check name сookies, i have slight doubt that in one of 3 you use cyrillic "c". in that case you will get ElementNotFound. Do not ask me how i get such thought)

app/(logged_in)/privacy-policy/page.test.tsx
DenisGordProgrammer, line 1: btw 11 same exact jest.mock. Unite them. like:
```
const MOCK_COMPONENTS = ['ContactUs', 'DataRetention', ...];
MOCK_COMPONENTS.forEach(name => {
  jest.mock(`~/shared/.../${name}`, () => ({
    [name]: () => <div data-testid={kebabCase(name)}>{name}</div>
  }));
});
```

app/(logged_in)/privacy-policy/page.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { DataRetention } from '~/shared/components/privacy-policy/data-retention/DataRetention';
+import { DataUsage } from '~/shared/components/privacy-policy/data-usage/DataUsage';
+import { DataWeCollect } from '~/shared/components/privacy-policy/data-we-collect/DataWeCollect';
+import { GoogleAuth } from '~/shared/components/privacy-policy/google-auth/GoogleAuth';
+import { IntroSection } from '~/shared/components/privacy-policy/intro-section/IntroSection';
+import { NewsletterSubscription } from '~/shared/components/privacy-policy/newsletter-subscription/NewsletterSubscription';
+import { SocialNetworks } from '~/shared/components/privacy-policy/social-networks/SocialNetworks';
+import { TargetedAds } from '~/shared/components/privacy-policy/targeted-ads/TargetedAds';
+import { UserRights } from '~/shared/components/privacy-policy/user-rights/UserRights';
+import { Cookies } from '~/shared/components/privacy-policy/сookies/Cookies';
```
DenisGordProgrammer: have another doubt that there we have same "с" (U+0441). in Linux/macOS there will cause import errors

DenisGordProgrammer: Image

app/constants/pageBlocks.ts
```diff
@@ -9,5 +10,16 @@ export const BLOCK_IDS = {
...
+  DATA_USAGE: 'DataUsage',
+  COOKIES: 'Cookies',
+  GOOGLE_AUTH: 'GoogleAuth',
+  SOCIAL_NETWORKS: 'SocialNetworks',
+  TARGETED_ADS: 'TargetedAds',
+  NEWSLETTER_SUBSCRIPTION: 'NewsletterSubscription',
+  DATA_RETENTION: 'DataRetention',
+  USER_RIGHTS: 'UserRights',
+  CONTACT_US: 'ContactUs',
+  PRIVACY_INTRO_SECTION: 'IntroSection'
```
DenisGordProgrammer, line 24: as url standarts better use kebab-case for values....but you know, its just to expand knowledge, im not really want push you renaming that in across all project

app/shared/components/configurable-list/__mocks__/ConfigurableList.tsx
DenisGordProgrammer, line 1: in the MockConfigurableListProps<T> interface, the onChange prop is defined as (item: T) => void. however, in the component itself, you use it for each item in items.map. 
well that means that the onChange function must take an argument that identifies which item has changed, or the outer state must be structured so that a single onChange function processes the entire array. 
currently, the onChange implementation in renderItem simply overwrites the object, but you don't pass in an index or ID, which makes it impossible to update the array correctly in the parent component.

my suggestion: Change the signature of onChange to (id: string, updatedItem: T) => void or (updatedItem: T) => void (if the function takes ID into account internally) so that the parent component can find the item by item.id.

app/shared/components/privacy-policy/__mocks__/setup-mocks.ts
```diff
@@ -0,0 +1,25 @@
...
+  useStore: (selector: (state: { readonly locale: 'uk'; readonly setField: typeof setFieldMock }) => unknown) =>
+    selector({ locale: 'uk', setField: setFieldMock })
+}));
+
+jest.mock('~/ds-components/collapsible-block/CollapsibleBlock');
+jest.mock('~/ds-components/text-field/TextField');
+jest.mock('~/shared/components/privacy-policy/components/points-list/PointsList');
+
+beforeAll(() => {
+  crypto.randomUUID = jest.fn(() => 'uuid-1') as typeof crypto.randomUUID;
```
DenisGordProgrammer: you redefine crypto.randomUUID here. well while that works for tests...crypto still is a global object, and it could cause test pollution, you know. my suggestion to use:
jest.spyOn(crypto, 'randomUUID').mockReturnValue('uuid-1')
and add afterAll(() => jest.restoreAllMocks()) (or, mhm, jest.clearAllMocks() in afterEach)

app/shared/components/privacy-policy/__mocks__/setup-mocks.ts
```diff
@@ -0,0 +1,25 @@
...
+export const mockRemovePoint = jest.fn();
+export const mockUpdatePoint = jest.fn();
+
+jest.mock('~/shared/hooks/use-page-block/usePageBlock', () => ({
+  usePageBlock: () => usePageBlockMock()
+}));
+
+jest.mock('~/store', () => ({
+  useStore: (selector: (state: { readonly locale: 'uk'; readonly setField: typeof setFieldMock }) => unknown) =>
+    selector({ locale: 'uk', setField: setFieldMock })
```
DenisGordProgrammer: also you hardly use only locale uk, but what if we need to test en locale?. suggestion to change mock in that way to it could work dynamically, OR use jest.spyOn, if possible.
in that specific way, if you want use ESPECIALLY jest.mock, add an opportunity to change locale via some helping value, that mock will use. Hope i explain it well

app/shared/components/privacy-policy/components/edit-description-list-note-block/EditDescriptionListNoteBlock.tsx
DenisGordProgrammer, line 1: here in EditDescriptionListNoteBlock component, you use listFieldName to access the array via block[listFieldName]. 
however, in the usePointsList hook, you pass the blockId (which is the block key in the BlocksMap), but you DO NOT pass information about which field in that block is a list. 
this will cause usePointsList to not be able to update the data in the page correctly (setField requires fieldName to update a specific field, not the entire block).

app/shared/components/privacy-policy/components/edit-description-list-note-block/EditDescriptionListNoteBlock.tsx
```diff
@@ -0,0 +1,92 @@
...
+      {descriptionBlock && (
+        <CustomTextField
+          fieldType="formatting"
+          title="Вступний текст секції"
+          value={descriptionBlock[currentLocale]}
+          onChange={(value) => handleTextChange('description', value)}
+        />
+      )}
+
+      {points.length > 0 && (
```
DenisGordProgrammer, line 74: in such situation if the list becomes empty (the user removes all points), the PointsList component will disappear from the DOM entirely. this will make it impossible to add new points (if the Add button is inside the PointsList) because the user will lose the interaction interface 
well i could get wrong here, so just double check behaivour

app/shared/components/privacy-policy/components/edit-paragraphs-block/EditParagraphsBlock.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+  runCommonBlockTests({
+    Component: () => <EditParagraphsBlock blockId={mockBlockId} title={mockTitle} />,
+    mockBlock: createStandardMockBlock().block,
+    checkParagraph: true,
+  });
+  it('should render a correct title if provided', () => {
+    usePageBlockMock.mockReturnValue({ block: createStandardMockBlock().block });
+    
+    render(<EditParagraphsBlock blockId={mockBlockId} title={mockTitle} />);
+  });
```
DenisGordProgrammer, line 24: well here we have it we have render, but dont have expect...it will works just like empty catch. moreover you dont check if usePageBlockMock use with right props. suggestion:
always use atleast one expect

app/shared/components/privacy-policy/components/edit-paragraphs-block/EditParagraphsBlock.tsx
DenisGordProgrammer, line 1: here you use useRef to store stableIds that are generated once during initialization. If the paragraphs array changes (for example, a paragraph is added or removed through another service or a language change), the stableIds will not be synchronized with the new list. 
this will lead to incorrect rendering (React will not understand which component has been updated) or errors in the CustomTextField. suggestion: 
stableIds should be derived from state OR each paragraph should have a unique ID that comes from the database, rather than being generated "here for second" via crypto.randomUUID in useRef

app/shared/components/privacy-policy/components/edit-paragraphs-block/EditParagraphsBlock.tsx
```diff
@@ -0,0 +1,74 @@
...
+      [currentLocale]: {
+        ...oldBlockDescription[currentLocale],
+        content: currentContentArray
+      }
+    };
+
+    setField(pageId, blockId, 'description', newDescription);
+  };
+
+  if (!paragraphs || paragraphs.length === 0) return null;
```
DenisGordProgrammer, line 56: also here we have 3 if's
if (!block) return <EditBlockSkeleton />

if (!('description' in block)) return null

if (!paragraphs || paragraphs.length === 0) return null

but, if the block exists but the paragraphs are empty, the component silently disappears from the interface. The user sees neither the Skeleton nor the block itself, which creates a "hole" in the UI. This is critical if the administrator has the ability to fill the empty block but once more, !just double-check!

app/shared/components/privacy-policy/components/edit-paragraphs-block/EditParagraphsBlock.tsx
DenisGordProgrammer, line 1: and also, i could be completely wrong, but i will say what im thinking about...
here we MABY could get Potential Race Condition on setField.
so in onParagraphChange you get oldBlockDescription from props (via usePageBlock). If the user types quickly, setField MAY NOT have time to update the block, and the next call to onParagraphChange will work with the "old" block object. suggestion: 
setField in your store must support functional updating (updater function: (prev) => newState) or you must ensure that the store update is atomic
once more !FEEL FREE TO IGNORE AFTER DOUBLE CHECKING!

app/shared/components/privacy-policy/components/points-list/__mocks__/PointsList.tsx
```diff
@@ -0,0 +1,38 @@
+import { JSONContent } from '@tiptap/react';
+import React from 'react';
+
+import { CustomTextField } from '~/shared/components/design-system/text-field/TextField';
+
+interface MockPoinstsListProps<T> {
+  points: T[];
+  addPoint: () => T;
```
DenisGordProgrammer, line 8: here the MockPoinstsListProps interface, the addPoint method is defined as () => T. BUT in the component, it is called as onClick={addPoint} (as a function with no arguments that returns nothing). If addPoint actually returns a new T element, then you ignore the execution result and the list state is not updated. suggestion: 
If addPoint is a function that should change state, it should be () => void or () => Promise<void>. If it creates an element and adds it to the state, the logic should be moved to the parent component, and the updated list should be passed there.
once more !FEEL FREE TO IGNORE AFTER DOUBLE CHECKING!

app/shared/components/privacy-policy/components/points-list/PointsList.test.tsx
```diff
@@ -0,0 +1,42 @@
+import { render, screen } from '@testing-library/react';
+import React from 'react';
+
+import { PointsList, PointsListProps } from './PointsList';
+import { ListPoint } from '~/shared/hooks/use-points-list/usePointsList';
+
+
+jest.mock('~/components/configurable-list/ConfigurableList');
```
DenisGordProgrammer: here PointsList uses a ConfigurableList (according to jest.mock('~/components/configurable-list/ConfigurableList')), but you are testing PointsList as an independent component. well if PointsList renders a ConfigurableList inside itself, then screen.getByText('Points:') and other elements will not be found because they are rendered inside a mocked component, which in tests returns undefined or null. Tests with getByText will fail.

app/shared/components/privacy-policy/components/points-list/PointsList.tsx
DenisGordProgrammer, line 1: in your PointsListProps interface, the addPoint method is defined as () => ListPoint, but you pass it to the ConfigurableList in the onCreate prop. according to the previous definition of ConfigurableList, onCreate has type () => void. and if addPoint returns a ListPoint and ConfigurableList does not expect it, the result of addPoint is simply lost and the state is not updated. moreover, the onChange type in ConfigurableList (which you use for updatePoint) expects an argument, and you pass updatePoint directly. BUT I COULD BE MISSING SOMETHING

app/shared/components/privacy-policy/data-usage/DataUsage.tsx
```diff
@@ -0,0 +1,60 @@
...
+  return (
+    <CollapsibleBlock title="Як ми використовуємо ваші дані">
+      <CustomTextField
+        fieldType="formatting"
+        title="Вступний текст секції"
+        value={block.title[currentLocale]}
+        onChange={(value) => handleChangeTitleText(value)}
+      />
+
+      {points.length > 0 && (
```
DenisGordProgrammer, line 54: we saw it before yeah, user deletes last element-pointlist disappears-user will not have opportunity click Add. suggestion:
always render the PointsList (or a separate container with an "Add" button), regardless of the presence of points, or move the add button higher in the hierarchy so that it is always available.

app/shared/components/privacy-policy/data-we-collect/DataWeCollect.test.tsx
```diff
@@ -0,0 +1,35 @@
+
+import { usePointsListMock } from '../__mocks__/setup-mocks';
+import { createStandardMockBlock, runCommonBlockTests } from '../test-utils/block-test-factory';
+import { DataWeCollect } from './DataWeCollect';
+import { createDocNode } from '~/__mocks__/utils';
+import { DataWeCollectBlock } from '~/types/store/pages/privacy-policy';
+
+
+jest.mock('~/components/configurable-list/ConfigurableList');
```
DenisGordProgrammer, line 9: DO U BLE CH E CK :D
you import usePointsListMock from setup-mocks, but in runCommonBlockTests you pass it in via the usePointsListMock prop. re-check that inside block-test-factory.ts this mock actually replaces the real usePointsList via jest.mock. If the test factory inside doesn't do jest.mock('~/shared/hooks/use-points-list/usePointsList', () => ({ usePointsList: usePointsListMock })), then passing the mock as an argument will be useless and the test will run with the real hook (or throw an error)

app/shared/components/privacy-policy/data-we-collect/DataWeCollect.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const blockId = BLOCK_IDS.DATA_WE_COLLECT;
+
+  const { block } = usePageBlock(pageId, blockId);
+  const currentLocale = useStore((state) => state.locale);
+  const setField = useStore((state) => state.setField);
+
+  const rawSections = block?.sections || [];
+  const sectionsList: DataWeCollectItemWithId[] = ensureIds(rawSections);
+
+  const { sections, addListPoint, removeListPoint, updateListPoint } = useSectionList({ blockId, pageId, sectionsList, currentLocale });
```
DenisGordProgrammer: here you are using sectionsList (obtained via ensureIds(rawSections)) in the handleChangeSectionSubtitle method, but you are getting the data to render from the sections hook (via useSectionList). if the useSectionList hook internally transforms the data or has a delay in updating, you will be updating the state of sectionsList, which is no longer the "source of truth" for the display. suggestion: 
all update logic should go exclusively through the useSectionList hook (e.g., via the updateSectionSubtitle method), rather than directly manipulating sectionsList

app/shared/components/privacy-policy/data-we-collect/DataWeCollect.tsx
```diff
@@ -0,0 +1,106 @@
...
+        sections.map((section, index) => (
+          <Box key={section.id}>
+            <CustomTextField
+              fieldType="formatting"
+              title={`Список ${index + 1}`}
+              value={section.title}
+              onChange={(value) => handleChangeSectionSubtitle(section.id, value)}
+            />
+            <ConfigurableList<LocalizedJSON & {
+              id: string;
```
DenisGordProgrammer, line 67: here in ConfigurableList<LocalizedJSON & { id: string; }> you pass items={section.points}. if section.points inside the useSectionList hook has a different structure than ConfigurableList expects (e.g. Tiptap JSON format), you will get a runtime error when trying to access item[currentLocale]. 
given that LocalizedJSON is usually an object of { uk: JSONContent, en: JSONContent }, attempting to access item[currentLocale] inside renderItem looks like attempting to access a property of an object that may not be there (since item should already be a localized content object)

app/shared/components/privacy-policy/intro-section/IntroSection.test.tsx
```diff
@@ -0,0 +1,41 @@
+import { render, screen } from '@testing-library/react';
+
+import { usePageBlockMock } from '../__mocks__/setup-mocks';
+import { runCommonBlockTests } from '../test-utils/block-test-factory';
+import { IntroSection } from './IntroSection';
+import { createDocNode } from '~/__mocks__/utils';
+
+beforeAll(() => {
+  let counter = 0;
+  crypto.randomUUID = jest.fn(() => `uuid-${++counter}`) as typeof crypto.randomUUID;
```
DenisGordProgrammer: once more about crypto:
use jest.spyOn(crypto, 'randomUUID').mockReturnValue('fixed-uuid') in beforeEach and 
jest.restoreAllMocks() in afterEach.

app/shared/components/privacy-policy/test-utils/block-test-factory.tsx
```diff
@@ -0,0 +1,130 @@
...
+    title: mockTitleJson,
+    description: mockDescriptionJson,
+    list: mockListPointJson,
+    note: mockNoteJson
+  }
+});
+
+interface CommonTestProps {
+  Component: React.ElementType;
+  mockBlock: unknown;
```
DenisGordProgrammer: here you are using the unknown type for mockBlock. and this means that when passing a mock from a specific test file (e.g. DataWeCollect), you have no guarantee that the object contains the fields needed for rendering (description, note, etc.). when running tests, you may get undefined errors when accessing internal fields in the components that the factory is trying to "test". suggestion: 
define a base interface BaseBlock with the fields that your factory expects, and use mockBlock: BaseBlock instead of unknown

app/shared/components/privacy-policy/cookies/Cookies.tsx
DenisGordProgrammer, line 1: you know what?))))) i have reason to say that you may use "с" (U+0441) in path /privacy-policy/[HERE]ookies

DenisGordProgrammer: im about name of folder, not the code itself btw

app/shared/hooks/use-points-list/usePointsList.test.tsx
```diff
@@ -0,0 +1,82 @@
...
+const defaultMockedProps = {
+  blockId,
+  list: initialList,
+  setField: mockSetField,
+  currentLocale,
+  pageId
+};
+
+beforeAll(() => {
+  crypto.randomUUID = jest.fn(() => 'uuid-1') as typeof crypto.randomUUID;
```
DenisGordProgrammer: you know what to do)

app/shared/hooks/use-section-list/useSectionList.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { emptyDoc } from '../use-points-list/usePointsList';
+import { useSectionList } from './useSectionList';
+import { ensureIds } from '~/lib/utils/ensureIds';
+import { setFieldMock } from '~/shared/components/privacy-policy/__mocks__/setup-mocks';
+
+Object.defineProperty(globalThis, 'crypto', {
+  value: {
+    randomUUID: jest.fn(() => 'uuid-1'),
+  },
+});
```
DenisGordProgrammer: you are overriding crypto.randomUUID via Object.defineProperty(globalThis, 'crypto', ...). this is a "hard" intervention in the Node.js/JSDOM environment. 
if other tests (or Jest itself) use crypto for internal purposes, this can lead to unexpected crashes. Also, you are not restoring the original state of crypto after the tests. suggestion:
use jest.spyOn(globalThis.crypto, 'randomUUID').mockReturnValue('uuid-1') and call jest.restoreAllMocks() in afterEach.

app/shared/components/privacy-policy/cookies/Cookies.tsx
DenisGordProgrammer: Image

DenisGordProgrammer (changes requested):

app/shared/components/privacy-policy/components/edit-paragraphs-block/EditParagraphsBlock.test.tsx
telare: yea, I missed this. I'll add the check

app/shared/components/privacy-policy/components/edit-description-list-note-block/EditDescriptionListNoteBlock.tsx
telare: yep, good point. But currently, an admin cannot remove all points from a list (only the first can be deleted). 
image

I think this should be reconsidered during or after the redesign task.

app/(logged_in)/privacy-policy/page.test.tsx
telare: Btw, both on my Ubuntu (laptop) and pc (Win10), all tests pass and no linting issues. I will recheck this.

app/(logged_in)/privacy-policy/page.tsx
telare: ok, I'll recheck it

app/shared/components/privacy-policy/data-usage/DataUsage.tsx
telare: yep, answered in one of the previous comments

app/(logged_in)/privacy-policy/page.test.tsx
telare: I have tried many variants to solve this, but I got errors due to jest.mock hoisting. The only way to make it like this: 

```

const mockComponents = [
  'ContactUs', 'DataRetention', 'DataUsage', 'DataWeCollect', 
  'GoogleAuth', 'IntroSection', 'NewsletterSubscription', 
  'SocialNetworks', 'TargetedAds', 'UserRights', 'Cookies'
];

jest.doMock('~/shared/components/privacy-policy', () => {
  const mockedModule: Record<string, any> = {};
  
  mockComponents.forEach(name => {
    mockedModule[name] = () => <div data-testid={mockKebabCase(name)}>{name}</div>;
  });
  
  return mockedModule;
});
 // eslint-disable-next-line @typescript-eslint/no-require-imports
const { default: Page } = require('./page');
```

And yeah, require is needed here, but I think it's better to use as it was.

app/shared/hooks/use-section-list/useSectionList.test.tsx
telare: good, I'll refactor this

app/(logged_in)/privacy-policy/page.tsx
telare: Yes, you are right. I have solved it in an upcoming commit.

app/(logged_in)/privacy-policy/page.test.tsx
DenisGordProgrammer: yeah, lint will not fall, problem much deeper, so just rename it)

app/shared/components/privacy-policy/components/edit-paragraphs-block/EditParagraphsBlock.tsx
telare: I see that in the current implementation, there is always at least one item in the content array from BE, so this could be a bug in the future if smth changes in the data from BE. 

But now I think it is good to take this into consideration, not more, maybe talk with designers to understand what can be visible in this case.

Thanks for pinpointing this!

app/shared/components/privacy-policy/__mocks__/setup-mocks.ts
telare: ok, I'll redo this

app/shared/components/privacy-policy/components/edit-description-list-note-block/EditDescriptionListNoteBlock.tsx
telare: I think for now, it is ok not to pass listFieldname cause in the db we have always 'list' as key to the points list, but for flexibility, it should accept as you said. 

`setField(pageId, blockId, 'list' as keyof BlocksMap[K]`
  
I will recheck my implementation.

app/shared/components/privacy-policy/components/points-list/PointsList.tsx
telare: Good point, and I have an explanation for this: 

If we pass a type to <ConfigurableList/> like this: <ConfigurableList<TeamMemberWithId> (or simply pass correctly typed items), it internally changes the type of onCreate from () => void to () => TeamMemberWithId, so I get an error if I change the return type of addPoint (for example). 

And yeah, this sounds interesting, but everything works all right, for now. 

I hope I have clarified that well.

app/shared/hooks/use-section-list/useSectionList.test.tsx
telare: I have investigated crypto module mocking in jest, especially in our project, and found interesting details. 

- First of all, I tried your approach, and it didn't work for me. 
- It must work because we are using 29.7.0 v of jest and the crypto module should be defined in the environment
- The reason is that we do have a crypto module in our JSDOM environment defined, BUT it doesn't contain any keys. Here is proof: 
```
console.log('typeof globalThis.crypto:', typeof globalThis.crypto); // object
console.log('Available keys:', Object.getOwnPropertyNames(globalThis.crypto)); // []
console.log('typeof randomUUID:', typeof globalThis.crypto.randomUUID); // undefined
```
- And because of this, whenever I tried to spy on the randomUUID method, I got an error. 
- So my code worked well, because I simply added this property for the 'globalThis' object with my mocked value via crypto.randomUUID or Object.defineProperty(globalThis, 'crypto'). However, you are absolutely right about the risk of test pollution.
- I tried many ways to solve this, and only that worked for me: 

 ```
const MOCK_UUID = 'uuid-1';
 let originalCrypto: any;
beforeAll(() => {
   originalCrypto = globalThis.crypto;
 
  Object.defineProperty(globalThis, 'crypto', {
     value: {
      ...originalCrypto,
       randomUUID: jest.fn().mockReturnValue(MOCK_UUID), 
     },
     configurable: true, 
   });
 });

 afterAll(() => {
 if (originalCrypto) {
   Object.defineProperty(globalThis, 'crypto', {
     value: originalCrypto,
      configurable: true,
   });
   } else {
     delete (globalThis as any).crypto;
   }
 })

```

So, here I am using my approach, but with correct cleanup (as I understood, I may be wrong). This approach doesn't pollute other tests, preserves the original value, and mocks the needed behavior. 

And I appreciate pinpointing this too, because I have gained more knowledge about Jest rn, thanks!

telare: Also, I see that a previously incorrect approach (with possible test pollution caused by mocking of the crypto module) was used in many files. So, I think it is better to create a separate task for refactoring this (Maybe, I'll create this next week)

app/shared/components/privacy-policy/__mocks__/setup-mocks.ts
telare: Ok, I take this into consideration. This part will be updated in the next commit

app/shared/components/privacy-policy/data-we-collect/DataWeCollect.tsx
telare: Fixed that in the new commit. It was interesting to analyze and understand, but now it is better.

app/shared/components/privacy-policy/data-we-collect/DataWeCollect.tsx
telare: I think in this way: 
Sections in useSectionList hook has type 
    id: string;
    title: JSONContent;
    points: (LocalizedJSON & {
        id: string;
    })[];
}[]

So, the only thing we need to ensure is that the data for this block from BE has a valid type, especially a 'list' field. So theoretically, as long as the BE payload is valid, this is safe.

And, yeah, I have understood your point and added a fallback for that case.

app/shared/components/privacy-policy/data-we-collect/DataWeCollect.test.tsx
telare: Rechecked, I accidentally mocked not what I meant, which caused the real hook (useSectionList) to run instead of a mocked one.
Solved.

app/shared/components/privacy-policy/components/edit-paragraphs-block/EditParagraphsBlock.tsx
telare: Good point. I think, for now, it can be as is, since we don't have a feature to add or remove a paragraph, and, as long as I consider it, there will be no case (or rare) where 2 admins will edit the same paragraphs in one block simultaneously.

app/shared/components/configurable-list/__mocks__/ConfigurableList.tsx
telare: - As I understand it, the onChange handler receives the entire new item, which already includes the id field. 
- So, a parent, in our case, the ConfigurableList component simply doesn't care about the item IDs to update. 
- We take into account that the id field is directly and only in a handler, for example, in FoundationFounders.tsx, we have updateMember method with updated item, which already contains the necessary id, so that when we pass this handler to the ConfigurableList in onChange, we already assume that the handler works with the id. 

I wish I explained this well, if not, lemme know)

app/shared/components/privacy-policy/components/points-list/__mocks__/PointsList.tsx
telare: The state actually updates perfectly fine right now because setPoints handles the side effect. While the click handler didn't use the returned element, the ConfigurableList interface required onCreate to return T.
I tested the behavior of ConfigurableList and related components after removing the return statement from the addPoint handler, and finally, everything worked. So, fixed, thanks.

app/shared/components/privacy-policy/components/points-list/PointsList.test.tsx
telare: The mock file exists and returns real JSX — it's not null or undefined. `jest.mock('~/shared/components/configurable-list/ConfigurableList')` with a manual mock file replaces the module with our custom implementation that renders an actual mocked markup with a button. 

So `screen.getByText('Пункти:')` works because it's rendered by PointsList itself (via <Typography>), and `screen.getByText('Додати пункт')` works because the mock renders `<button>{addBtnLabel}</button>` with the prop passed from PointsList.

The only real issue was the wrong path in jest.mock — ~/components/... instead of ~/shared/components/...

DenisGordProgrammer (approved):
very good

yuliiayarova (approved):

PR #654, opened 2026-07-01
fix(files): prevent file details panel from deforming cards size 

### LINKED ISSUE #653: (SP: 2) [Bug][Files] File details panel deforms the file cards
## Description

On the Files page, after selecting a file, the file details panel should open above the file list without affecting the page layout.

### Currently:

* no file is selected by default; file details are displayed as part of the page, causing file cards to become smaller;
* opening the details panel changes (deforms) the cards;
* the selected file state does not fully match the approved design;
* pressing the Esc key does not close the details panel.

Image

### Expected result

* When the Files page is opened, no file is selected or highlighted by default.
* Once a file is selected, it is highlighted according to the approved design.
* Opening the details panel must not change or deform the size of the file cards.
* File cards must retain the same size as before the details panel is opened.
* Users can select another file while the details panel remains open.
* Only one file can be selected at a time.
* Pressing the Esc key closes the details panel.
* Clicking the close (X) button closes the details panel.
* If a user selects a file from the middle of the list, the selected card should remain visible after the details panel is opened.

### Notes

* The behavior should be consistent across all tabs: All, Images, Docs, Audio, and Favorites.
* The implementation should match the approved design.

## Acceptance criteria

1. Open the Files page.
2. Verify that no file is selected by default.
3. Select any file.
4. Verify that opening the details panel does not resize the file cards.
5. Try selecting another file while the details panel is open.
6. Press the Esc key.
7. Open the details panel again and click the close (X) button.

## Expected Result

Image

Halyna-Trush:
## Description

Fixed the behavior of the file details panel on the Files page to match the approved design.

Closes #653 
Changes:

* The file details panel now opens without affecting the page layout.
* Prevented file cards from resizing when the details panel is opened.
* Updated the selected file state to match the approved design.
* Ensured that only one file can be selected at a time.
* Added support for closing the details panel using the Esc key and the Close (X) button.
* Preserved the scroll position so the selected file remains visible after opening the details panel.
* Applied consistent behavior across all Files tabs: All, Images, Docs, Audio, and Favorites.

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Open the Files page.
2. Verify that no file is selected by default.
3. Select any file.
4. Verify that opening the details panel does not resize the file cards.
5. Select another file while the details panel is open and verify that the selection updates correctly.
6. Press Esc and verify that the details panel closes.
7. Open the details panel again and close it using the Close (X) button.
8. Select a file from the middle of the list and verify that it remains visible after the details panel is opened.
9. Repeat the verification in the All, Images, Docs, Audio, and Favorites tabs.

## Video / Screenshots

https://github.com/user-attachments/assets/7cf233e0-2ffb-46f5-b0d1-4fe1da55e3e2

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues

---

### DIFF
--- a/app/(logged_in)/files/FilesPageContent.tsx
+++ b/app/(logged_in)/files/FilesPageContent.tsx
@@ -2,7 +2,7 @@
 
 import { Box, Button } from '@mui/material';
 import { Upload } from 'lucide-react';
-import { useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import toast from 'react-hot-toast';
 
 import { getFilePageContentWrapper, styles } from './FilesPageContent.styles';
@@ -166,7 +166,7 @@ const usageToLink = (pageId?: string | null) => {
 export function FilesPageContent({ activeTab }: FilesPageContentProps) {
   const [view, setView] = useState<FilesCardsLayoutView>('grid');
   const [selectedFileId, setSelectedFileId] = useState<string | null>(null);
-  const [hasInitializedSelection, setHasInitializedSelection] = useState(false);
+  const fileCardRefs = useRef<Record<string, HTMLDivElement | null>>({});
   const [isUploadModalOpen, setIsUploadModalOpen] = useState(false);
   const [deleteModalState, setDeleteModalState] = useState<{ open: boolean; fileId: string | null }>({
     open: false,
@@ -181,6 +181,14 @@ export function FilesPageContent({ activeTab }: FilesPageContentProps) {
   const { data, loading, error, refetch } = useAllAssets();
   const [createAsset] = useCreateAssetMutation();
   const [deleteAsset, { loading: isDeleting }] = useDeleteAssetMutation();
+
+  const handleItemRef = useCallback((itemId: string, node: HTMLDivElement | null) => {
+    fileCardRefs.current[itemId] = node;
+  }, []);
+
+  const handleItemClick = useCallback((item: FilesCardsLayoutItem) => {
+    setSelectedFileId(item.id);
+  }, []);
 
   const handleOpenUploadFlow = () => {
     setUploadModalInitial({ tab: 'UPLOAD' });
@@ -285,8 +293,9 @@ export function FilesPageContent({ activeTab }: FilesPageContentProps) {
       toast.success('Файл успішно видалено');
       setDeleteModalState({ open: false, fileId: null });
       if (selectedFileId === id) setSelectedFileId(null);
-    } catch (error: any) {
-      toast.error(error.message || 'Не вдалося видалити файл. Спробуйте пізніше.');
+    } catch (error: unknown) {
+      const message = error instanceof Error ? error.message : 'Не вдалося видалити файл. Спробуйте пізніше.';
+      toast.error(message);
     }
   };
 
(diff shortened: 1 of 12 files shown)


app/(logged_in)/files/FilesPageContent.tsx
```diff
@@ -310,20 +319,49 @@ export function FilesPageContent({ activeTab }: FilesPageContentProps) {
...
+  }, [filteredFiles, selectedFileId]);
+
+  useEffect(() => {
+    if (!selectedFileId) {
       return;
     }
 
-    if (selectedFileId && !filteredFiles.some((file) => file.id === selectedFileId)) {
-      setSelectedFileId(filteredFiles[0].id);
+    const animationFrameId = globalThis.requestAnimationFrame(() => {
```
telare: Suggestion: re-test this behavior without `requestAnimationFrame`. Since `useEffect `runs after the browser has painted (after `useLayoutEffect`) and the layout geometry is fully calculated, using `requestAnimationFrame `might be redundant here.

app/(logged_in)/files/FilesPageContent.tsx
```diff
@@ -182,6 +182,14 @@ export function FilesPageContent({ activeTab }: FilesPageContentProps) {
   const [createAsset] = useCreateAssetMutation();
   const [deleteAsset, { loading: isDeleting }] = useDeleteAssetMutation();
 
+  const handleItemRef = useCallback((itemId: string, node: HTMLDivElement | null) => {
+    fileCardRefs.current[itemId] = node;
+  }, []);
+
+  const handleItemClick = useCallback((item: FilesCardsLayoutItem) => {
```
telare: I suggest you recheck whether `useCallback `is necessary here. Because yes, you wanna keep a stable reference to this function that you pass to the `FilesCardsLayout`, but the `FilesCardsLayout `doesn't use `React.memo` to check props to avoid re-renders. 

So `useCallback `here can be redundant here. Also, as I know, we should memoize only the code whose calculations are so computationally intensive in React.

telare (changes requested):
Good work! No critical errors/bugs, only small adjustments can be applied/rechecked )

app/(logged_in)/files/FilesPageContent.tsx
DenisGordProgrammer: wel well well, what a great style of comments, "suggestion:" you say, ahhahaha, nice!

Halyna-Trush: Done. I re-tested this behavior without `requestAnimationFrame` and the selected card still remains visible after the details panel is opened.
Since this logic runs inside `useEffect`, the layout is already updated enough for `scrollIntoView` to work correctly here, so I removed `requestAnimationFrame` and call `scrollIntoView` directly.

app/(logged_in)/files/FilesPageContent.tsx
Halyna-Trush: Done. I rechecked this and agree that `useCallback` is redundant for `handleItemClick`, because `FilesCardsLayout` is not memoized with `React.memo`, so a stable callback reference does not prevent re-renders here.
I simplified `handleItemClick` to a regular function.
I kept `useCallback` only for `handleItemRef`, because it is used as a ref callback for storing card DOM nodes by file id.

telare (approved):

vladashvch (approved):

PR #701, opened 2026-07-17
Feature: implement server-side search, filtering, sorting and pagination for the /creativity page

### LINKED ISSUE #686: (SP: 3) Implement server-side pagination and filtering for /creativity Page
## Description

Refactor the /creativity page to move all data processing (pagination, status/language filtering, and text search) from the client to the backend. The current implementation fetches all records and filters them locally, which is not scalable.

## Acceptance criteria

1. Typing in the search bar must trigger an API request with the search query parameter.
2. Pagination must reflect the total count provided by the backend.
3. Memory usage on the client is minimized as only the current page of data is processed
4. The WorksTable should display the exact slice of data returned by the backend for the current page and active filters.

## Expected Result

The `/creativity` page becomes a "pure" UI component that displays data fetched directly from the server according to the active filters, search query, and current page.

## Extra

Design:
https://www.figma.com/design/IDQ5XmNo6PQBmBlL3DBm6Q/Admin-Liatoshynsky-Foundation?node-id=10800-91115&t=YhbQLNEVJjqLDKDd-0

Image

vladashvch:
## Description

Implement server-side filtering, search, sorting, and pagination for the `/creativity` page.

Previously, the frontend fetched all opuses, opus groups, and compositions in separate requests, merged the data on the client, and then applied filtering, searching, and sorting locally. This approach did not scale well and required unnecessary data processing in the browser.

The page now requests only the data required for the current tab, active filters, search query, sort order, and requested page from the backend.

## What was changed

- Moved filtering, search, sorting, and pagination logic from the frontend to the backend.
- Replaced multiple client-side requests with a single `usePaginatedWorks` query.
- Added backend pagination support while keeping the existing frontend `Pagination` component.
- Implemented tab-specific data retrieval:
  - `Всі` (all) tab – returns both opus groups and standalone compositions.
  - `Опуси` (op) tab – returns only `op` groups (`numberKind = op`).
  - `Безопусні` (woo) tab – returns only `woo` groups (`numberKind = woo`).
  - `Твори`(work) tab – returns only standalone compositions (without `opus_id`).
- Added dedicated tab handlers to encapsulate backend logic for each tab.
- Implemented mixed pagination logic for the `All` tab:
  - pages display opus groups first;
  - than it shows `bo` (woo) groups;
  - once opus groups are exhausted, standalone compositions are appended;
  - applied the same mixed-pagination approach for transitions between `Opuses → Without Opus` and `Without Opus → Works` without duplicating or skipping records.
- Added repository helpers to normalize legacy data by assigning default draft statuses to records that previously had no status in the database (both `Opus` and `Composition` repositories).
- Extended `buildBaseQuery` to support custom searchable fields instead of relying on a fixed set of fields.
- Updated and fixed unit tests to reflect the new server-side behavior.

> **Note:** Sorting by title will work correctly after the `number` field is migrated from a `string` to a `number`, as intended.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Open the `/creativity` page.
2. Verify that each tab displays the correct data:
   - `All` – opus groups first, followed by without-opus groups, then standalone compositions.
   - `Opuses` – only `op` groups.
   - `Without Opus` – only `woo` groups.
   - `Works` – only standalone compositions.
3. Verify that pagination displays the correct number of pages and loads the expected data.
4. Verify that changing pages triggers a new API request.
5. Verify that status and language filters trigger backend requests and return the expected results.
6. Verify that searching filters records by title and work number.
7. Verify that sorting works correctly for all supported fields.

### How it used to work

image

### How it works now
***Всі tab***
image

***Всі tab***: correct order of `op`, `bo` and `works`
image

***Всі tab***: all drafts with edge cases 
image

***Опуси tab***
image

***Безопусні tab***
image

***Твори tab***
image

### Tests

image
image
image
image
image
image
image
image
image

## Checklist

- [ ] Code compiles and runs correctly
- [ ] Tests pass successfully
- [ ] Linting checks are clean
- [ ] No Sonar issues
- [ ] Branch is up to date with the target branch

---
Closes #686

### DIFF
--- a/src/infrastructure/repositories/compositionRepository/compositionRepository.ts
+++ b/src/infrastructure/repositories/compositionRepository/compositionRepository.ts
@@ -1,10 +1,11 @@
 import mongoose, { Model } from 'mongoose';
 
 import { createBaseRepository } from '../baseRepository/baseRepository';
-import { buildBaseQuery, getBaseSort } from '../helpers';
+import { buildBaseQuery, combineConditions, fieldCondition, getBaseSort } from '../helpers';
 import { Composition } from '~/domain/entities/Composition';
 import { CompositionFilters, CompositionInput, ICompositionRepository } from '~/domain/repositories/compositionRepository';
 import dbConnect from '~/infrastructure/db/connect';
+import { OpusStatus } from '~/types/graphql/generated/graphql';
 
 export type DbComposition = {
   _id: { toString(): string };
@@ -49,9 +50,13 @@ export const CompositionRepository = ({ CompositionModel }: CompositionRepoDeps)
     model: CompositionModel,
     toEntity,
     buildQuery: (filters) => {
-      const query = buildBaseQuery(filters);
-      if (filters?.opusId) query.opusId = filters.opusId;
-      return query;
+      const query = buildBaseQuery({ ...filters, statuses: undefined }, ['genre', 'title.uk', 'title.en']) ?? {};
+
+      return combineConditions<DbComposition>([
+        query,
+        fieldCondition<DbComposition>('status', filters?.statuses as unknown as string[], OpusStatus.Draft as string),
+        filters?.opusId ? { opusId: filters.opusId } : null
+      ]);
     },
     getDefaultSort: getBaseSort
   });
--- a/src/infrastructure/repositories/compositionRepository/compositionRepository.ts
+++ b/src/infrastructure/repositories/compositionRepository/compositionRepository.ts
@@ -55,7 +55,8 @@ export const CompositionRepository = ({ CompositionModel }: CompositionRepoDeps)
       return combineConditions<DbComposition>([
         query,
         fieldCondition<DbComposition>('status', filters?.statuses as unknown as string[], OpusStatus.Draft as string),
-        filters?.opusId ? { opusId: filters.opusId } : null
+        filters?.isStandalone === true ? { opusId: null } : null,
+        filters?.isStandalone === false ? { opusId: { $ne: null } } : null
       ]);
     },
     getDefaultSort: getBaseSort
--- a/src/infrastructure/repositories/compositionRepository/compositionRepository.ts
+++ b/src/infrastructure/repositories/compositionRepository/compositionRepository.ts
@@ -50,7 +50,7 @@ export const CompositionRepository = ({ CompositionModel }: CompositionRepoDeps)
     toEntity,
     buildQuery: (filters) => {
       const query = buildBaseQuery(filters);
-      if (filters?.isStandalone) query.opusId = null;
+      if (filters?.opusId) query.opusId = filters.opusId;
       return query;
     },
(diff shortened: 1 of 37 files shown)


DenisGordProgrammer (changes requested):
Blocking issue found.

`src/infrastructure/repositories/compositionRepository/compositionRepository.ts`

`buildQuery` currently adds opusId condition with truthy check:
`filters?.opusId ? { opusId: filters.opusId } : null`

This removes the filter when `opusId` is `null`. But `null` is the explicit value used by `paginatedWorks` handlers to fetch standalone works (`opusId: null`).

Impact:
- Works tab can include grouped compositions.
- Mixed tab can leak grouped compositions into standalone list and break pagination consistency.

Please treat `opusId: null` as a valid explicit filter value (check property presence, not truthiness).

DenisGordProgrammer (changes requested):
Requesting changes due to a blocking bug in standalone works filtering. See inline comment in src/infrastructure/repositories/compositionRepository/compositionRepository.ts.

src/infrastructure/repositories/compositionRepository/compositionRepository.ts
```diff
@@ -49,9 +50,13 @@ export const CompositionRepository = ({ CompositionModel }: CompositionRepoDeps)
     model: CompositionModel,
     toEntity,
     buildQuery: (filters) => {
-      const query = buildBaseQuery(filters);
-      if (filters?.isStandalone) query.opusId = null;
-      return query;
+      const query = buildBaseQuery({ ...filters, statuses: undefined }, ['genre', 'title.uk', 'title.en']) ?? {};
+
+      return combineConditions<DbComposition>([
+        query,
```
DenisGordProgrammer, line 56: Could we switch this to an explicit null check? `opusId: null` is the standalone-works filter, but this truthy condition drops it. Suggestion: include the condition when the key is present, even if value is null.

vladashvch: You're absolutely right, thanks for catching this! 

I’ve updated the logic to use an isStandalone flag (boolean or null) instead of relying on implicit checks. So now:
- if the filter is explicitly set to true, it targets documents where opusId is null. 
- If set to false, it targets documents where opusId is present ($ne: null)
- If the filter is null, no condition is added, returning all compositions.

Thanks a lot for your suggestion and review!

DenisGordProgrammer: Good fix. The strict equality checks (`=== true` / `=== false`) correctly handle all three states: standalone filter, non-standalone filter, and no filter when null. Thanks for addressing it.

DenisGordProgrammer (approved):

annak413 (approved):

PR #702, opened 2026-07-17
feat(artistry): add artistry page layout, blocks and tests

### LINKED ISSUE #680: (SP: ) Implement drag & drop reodering for blocks of composition editor page
## Description

The goal of this task is to allow admins to reorder blocks within compositions on the composition editor page. 
Leverage existing logic for reordering that is used for /about-us & /privacy-policy pages (look for all layers of backend and frontend to grasp all logic). 

**IMPORTANT NOTE**: reordering of list items for "Твори в опусі" & "Пункти секції:" is out of scope for this task!

## Steps

1. Identify the page for editing a composition
2. Understand the logic behind the reordering of blocks using the logic that is defined for the about-us & privacy-policy pages blocks 
3. Update the database, Mongoose schemas, and GraphQL with reordering logic
4. Update UI using existing wrappers such as SortableBoard, SortableList, etc., and hooks for this
5. Cover all your changes with tests for both BE and FE with appropriate coverage (as much as u can) 

## Expected Result
- [ ] Admins can reorder blocks within the composition editor page 
- [ ] Changes to the block order are correctly applied to the database 

## Extra
[Figma reference](https://www.figma.com/design/IDQ5XmNo6PQBmBlL3DBm6Q/Admin-Liatoshynsky-Foundation?node-id=8396-90234&t=2qIE8SlPaUtuQT90-4)

varenichek22:
## Description

Implemented the management UI for the "Artistry" page. Created a new Mongoose schema with an `ArtistryPage` discriminator and added editable blocks: `TitleWithQuote` (for TipTap text fields) and `MusicTableSection`. Configured DnD reordering with strict layout locking for the first block. Written unit tests with 100% coverage.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

> !!! **Cross-repository Dependency:** Merge and deploy this PR **after** the client PR.

1. Run the admin panel locally (`npm run dev`) and navigate to the "Artistry" page.
2. Verify that `TitleWithQuote` and `MusicTableSection` edit blocks work correctly.
3. Try reordering blocks via Drag & Drop and ensure the first block remains locked at index 0.
4. Run tests to verify 100% coverage: `npx jest artistry`

## Video / Screenshots

image

image

image

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues

---

### DIFF
--- a/app/shared/components/artistry/MusicTableSection/MusicTableSection.tsx
+++ b/app/shared/components/artistry/MusicTableSection/MusicTableSection.tsx
@@ -1,0 +1,51 @@
+'use client';
+import ArrowForwardIcon from '@mui/icons-material/ArrowForward';
+import { Box, Button,Typography } from '@mui/material';
+import Link from 'next/link';
+
+import { BLOCK_IDS, PAGE_IDS } from '~/constants/pageBlocks';
+import CollapsibleBlock from '~/shared/components/design-system/collapsible-block/CollapsibleBlock';
+import { EditBlockSkeleton } from '~/shared/components/edit-block-skeleton/EditBlockSkeleton';
+import { usePageBlock } from '~/shared/hooks/use-page-block/usePageBlock';
+
+export const MusicTableSection = () => {
+  const pageId = PAGE_IDS.ARTISTRY;
+  const blockId = BLOCK_IDS.MUSIC_TABLE;
+
+  const { block } = usePageBlock(pageId, blockId);
+
+  if (!block) return <EditBlockSkeleton />;
+
+  return (
+    <CollapsibleBlock title="Таблиця музичних творів">
+      <Box
+        sx={{
+          p: 2,
+          mt: 1,
+          backgroundColor: 'action.hover',
+          borderRadius: 1,
+          display: 'flex',
+          flexDirection: 'column',
+          gap: 2
+        }}
+      >
+        <Typography variant="body2" color="text.secondary">
+          Цей блок динамічно завантажує таблицю композицій з бази даних. Додавання та редагування самих музичних творів відбувається у відповідному розділі адмін-панелі.
+        </Typography>
+
+        <Box>
+          <Button
+            component={Link}
+            href="/creativity"
+            variant="outlined"
+            color="primary"
+            size="small"
+            endIcon={<ArrowForwardIcon />}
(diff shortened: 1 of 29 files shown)


src/domain/entities/Opus.ts
```diff
@@ -64,4 +64,6 @@ export type Opus = {
 
   createdAt: string;
   updatedAt: string;
+
+  blocksOrder?: string[] | null;
```
telare, line 67: could u clarify why we need both null and undefined there?

telare (changes requested):

src/domain/entities/Opus.ts
DenisGordProgrammer: where is undefined huh? 
optional blocks order could be either array of strings, or null

DenisGordProgrammer (approved):

src/domain/entities/Opus.ts
telare: As I know, an optional property of an object implicitly gets `| undefined` added (unless `exactOptionalPropertyTypes: true` is set in tsconfig), so, possibly, we can expect blocksOrder: undefined, couldn't we?

varenichek22: The reason we explicitly need both undefined and null here is because: if the frontend doesn't send blocksOrder, it comes as undefined, and as for null, we need this because Mongoose might return null if the field wasn't initialized

app/shared/components/artistry/MusicTableSection/MusicTableSection.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const blockId = BLOCK_IDS.MUSIC_TABLE;
+
+  const { block } = usePageBlock(pageId, blockId);
+
+  if (!block) return <EditBlockSkeleton />;
+
+  return (
+    <CollapsibleBlock title="Таблиця музичних творів">
+      <Box
+        sx={{
```
telare, line 22: non-blocking, just to be clear, it is better to use a separate styles.ts file instead of inline

telare (approved):

PR #704, opened 2026-07-17
fix(auth): verify reset token via use-case instead of self-fetch to /api/graphql

### LINKED ISSUE #703: [Bug] [lf-admin] Password reset link shows "invalid link" error on production despite valid, non-expired token
**Environment:** OS, browser, it's version, device (for mobile devices).
**Reproducible:** always, rarely, sometimes etc.
**Build found:** last commit from https://github.com/ita-social-projects/GreenCityClient/tree/dev
E.g. "commit 147fa44"

**Preconditions**
On production, clicking the password reset link from the email always returns an "invalid link" error, even when the token is valid and not expired. This does not reproduce on staging.
Root cause: verifyToken() in app/reset-password/page.tsx makes a self-referencing HTTP fetch call from the server component to the app's own /api/graphql endpoint. On production, this internal request gets rejected with 401 Unauthorized (likely due to Azure/WAF/Front Door config blocking the request), causing the catch block to return false.

Actual result:
Reset password page always displays: "Посилання для відновлення пароля вже було використано або втратило чинність" - even for a fresh, valid token.

Expected result:
Valid, non-expired reset tokens should pass verification and show the reset password form.
Steps to reproduce:

dest411:
## Description
Refactored the token verification logic in the reset password page. Replaced the direct self-fetching HTTP request to `/api/graphql` with an invocation of `verifyResetTokenUseCase` via the root dependency injection container. This fixes an issue where password resetting worked on the staging environment but failed in production due to network/routing restrictions on direct internal API self-fetches.

## Type of change

- [ ] New feature
- [X] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Go to the password reset page with a valid token parameters (`/reset-password?token=YOUR_TOKEN`).
2. Verify that the system validates the token and correctly renders the change password form.

Closes: https://github.com/Liatoshynsky-Foundation/lf-admin/issues/703

### DIFF
--- a/app/reset-password/page.stories.tsx
+++ b/app/reset-password/page.stories.tsx
@@ -1,14 +1,18 @@
 import type { Meta, StoryObj } from '@storybook/react';
-import { expect, userEvent, within } from '@storybook/test';
+import { expect, fn, userEvent, within } from '@storybook/test';
 import { graphql, HttpResponse } from 'msw';
 
 import { getRouter } from '../../.storybook/mocks/next-navigation';
 import { withMswHandlers } from '../../.storybook/msw';
 import ResetPasswordPage from './page';
 
-const verifyTokenHandler = graphql.query('VerifyResetToken', () =>
-  HttpResponse.json({ data: { verifyResetToken: true } })
-);
+const mockExecute = fn();
+
+jest.mock('~/container/index', () => ({
+  createRootContainer: jest.fn(() => ({
+    resolve: jest.fn(() => ({ execute: mockExecute }))
+  }))
+}));
 
 const meta = {
   title: 'Pages/ResetPasswordPage',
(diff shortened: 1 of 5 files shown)


app/reset-password/page.stories.tsx
```diff
@@ -1,14 +1,18 @@
...
 
 import { getRouter } from '../../.storybook/mocks/next-navigation';
 import { withMswHandlers } from '../../.storybook/msw';
 import ResetPasswordPage from './page';
 
-const verifyTokenHandler = graphql.query('VerifyResetToken', () =>
-  HttpResponse.json({ data: { verifyResetToken: true } })
-);
+const mockExecute = fn();
+
```
DenisGordProgrammer: Non blocking: using `jest.mock` inside a Storybook story file can fail outside Jest runtime. Could we keep this mocked via MSW or Storybook-level module mocking so the story works in Storybook UI too?

DenisGordProgrammer (approved):

app/reset-password/page.stories.tsx
dest411: I'll change it👍🏻

telare (approved):

PR #712, opened 2026-07-19
feature: implement drag-and-drop sorting for performances within a group

### LINKED ISSUE #711: (SP: 2) Implement drag and drop reordering for performances within a group
## Description

Add support for existing drag-and-drop functionality on the group of performances page. Reuse existing components such as `SortableBoard`, `SortableContainer`, `SortableList`, `SortableItemWrapper`, and `Grip` together with the `handleSortableDragEnd` utility.

#### Performance ordering within a group

- The administrator can reorder performances within a group using drag-and-drop in the group edit view.
- The order of performances within a group is defined manually by the administrator.
- By default, performances are ordered by the sequence in which they were added to the group.

#### Persistence

- The saved order of performances within each group is reflected on the client side.

#### Data integrity

- Changing the order does not modify the content of records and does not affect linked files.

## Extra

Image

vladashvch:
## Description

Added drag-and-drop reordering support to the GroupPerformancesSection component.

### Changes

- Reused the existing `SortableList` component in `GroupPerformancesSection` to provide a consistent drag-and-drop experience.
- Reused the existing `handleSortableDragEnd` utility for performance reordering.
- Extracted the performance row into a dedicated `PerformanceRow` component.
- Wrapped each row in `SortableItemWrapper` for drag interactions.
- Increased test coverage 

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Check out this branch
2. Open the `/creativity` page
3. Create a new group via 'Створити' - 'Група'
4. In the 'Всі версії виконання опису' section, add several performances using the 'Додати пункт' button
5. Verify that performances can be reordered using drag-and-drop while creating the group
6. Save the group
7. Open the created group for editing and verify that performances can still be reordered
8. Save the changes and confirm that the performance order is preserved and displayed correctly

## Video / Screenshots

https://github.com/user-attachments/assets/3e5bcc29-b0cc-405a-be55-bc1fb6dd39d6

## Checklist

- [ ] Code compiles and runs correctly
- [ ] Tests pass successfully
- [ ] Linting checks are clean
- [ ] No Sonar issues
- [ ] Branch is up to date with the target branch

---
Closes #711

### DIFF
--- a/app/shared/components/creativity/group/performances-section/GroupPerformancesSection.tsx
+++ b/app/shared/components/creativity/group/performances-section/GroupPerformancesSection.tsx
@@ -4,7 +4,7 @@ import { useMemo, useState } from 'react';
 
 import { styles } from './GroupPerformancesSection.styles';
 import { PerformanceRow } from './PerformanceRow';
-import { GroupPerformance } from '~/constants/creativity';
+import { GroupPerformance, NormalizedGroupPerformance } from '~/constants/creativity';
 import { EditorLanguage } from '~/constants/publications';
 import { generateUniqueId } from '~/lib/utils/generateUniqueId';
 import { handleSortableDragEnd } from '~/lib/utils/sortableDragEndHelper';
@@ -71,6 +71,7 @@ export const GroupPerformancesSection = ({
   const langKey = currentLanguage === 'UA' ? 'uk' : 'en';
   const [performanceIdToDelete, setPerformanceIdToDelete] = useState<string | null>(null);
 
+  console.log('performances', performances);
   const preparedPerformances = useMemo(() => {
     return performances.map((item) => ({
       ...item,
@@ -79,7 +80,7 @@ export const GroupPerformancesSection = ({
   }, [performances]);
 
   const handleAddPerformance = () => {
-    const newPerformance: GroupPerformance = {
+    const newPerformance: NormalizedGroupPerformance = {
       id: generateUniqueId(),
       url: '',
       caption: { uk: '', en: '' }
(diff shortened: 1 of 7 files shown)


DenisGordProgrammer (changes requested):
Please address the inline comment before merge.

app/shared/components/creativity/group/performances-section/GroupPerformancesSection.tsx
```diff
@@ -116,52 +121,34 @@ export const GroupPerformancesSection = ({
...
-              </Box>
-
-              <IconButton onClick={() => setPerformanceIdToDelete(item.id || '')} sx={styles.actionIcon}>
-                <TrashIcon />
-              </IconButton>
-            </Box>
-          ))}
+          <SortableList
+            id="group-performances-list"
+            items={performances.map((item) => item.id)}
```
DenisGordProgrammer: `GroupPerformance.id` used to be optional, so legacy records can still arrive without an id. This new DnD path now passes ids directly into `SortableList`/`SortableItemWrapper` (`items={performances.map((item) => item.id)}`), and undefined ids will break or make sorting unstable for existing data. Please normalize missing ids before rendering (for example, map performances through a helper that injects `generateUniqueId()` when `id` is absent) so old records remain editable.

vladashvch: Yes, I agree. I’ve updated the code to normalize the data to handle cases where an ID might be missing

DenisGordProgrammer (approved):

app/shared/components/creativity/group/performances-section/GroupPerformancesSection.tsx
DenisGordProgrammer: Fix confirmed - useMemo normalizes missing ids before DnD. Non-blocking note: the fallback generateUniqueId() runs on every new performances reference, so if the parent frequently re-creates the array without interactions, ids could shift between renders. Low risk in practice since the first write-back persists the generated id - but normalizing at the data-loading layer would be more robust.

annak413 (approved):

PR #695, opened 2026-07-16
fix: creating publication meta title shorter

Fronik123:
## Description
Closes: #431 

Description: According to the requirements, the Meta title field is required and must contain at least 2 characters. However, the GraphQL API successfully creates a publication when the Ukrainian title value contains only 1 character. The invalid value is saved and displayed publicly as the H1 title on the publication page.

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

Postman

1. Execute the Login request in Postman.
2. Verify that authorization data is successfully saved in Postman variables.
3. Open the GraphQL request Mutations/Create News.
4. In GraphQL Variables, set title.uk to 1 character while keeping all other required fields valid:
```
{
   "input": {
     "adminTitle": "Test Yana",
     "title": {
       "uk": "Т",
       "en": "Test news"
     }
   }
}
```
5. Send the request.
6. Check the API response.

Actual result:
The API returns 200 OK.
The post was not created or update.

Admin panel

1. Go to the Admin Panel.
2. click on menu новина та події
3. Select any publication.
4 Click the three-dot menu and choose SEO налаштуваня
5. Change the title by one character and try to save it by clicking 'Опублікувати' or 'опублікувати та вийти'

Actual result:
A warning will appear stating that you must enter two characters.

## Video / Screenshots

Postman
telegram-cloud-photo-size-2-5386798234220565894-x

Admin panel
telegram-cloud-photo-size-2-5386798234220565901-y
telegram-cloud-photo-size-2-5386798234220565903-y

## Checklist

- [ ] Code compiles and runs correctly
- [ ] Tests pass successfully
- [ ] Linting checks are clean
- [ ] No Sonar issues
- [ ] Documentation updated (if applicable)
- [ ] All comments and requested changes addressed
- [ ] Branch is up to date with the target branch
- [ ] Linked issue/task included
- [ ] Task moved to the review column on the board

---

Issue #602, opened 2026-06-11
(SP: 5) Drag-and-drop reordering for blocks and list items on admin edit pages


telare:
## Description

Currently, there is no reordering logic & UI for an admin in /about-us or similar pages.
The goal is simple: allow an admin to reorder list items and blocks on an edit page.

**Note:** 
This task is related to this task: https://github.com/Liatoshynsky-Foundation/lf-admin/issues/271

## Steps

1. Install all necessary dependencies for reordering (for example, dnd kit)
2. Create reusable wrappers for an ordered list and list item with reordering logic
3. Apply all new components on existed pages (e.g. /about-us & /privacy-policy)
4. Ensure that a user can easily reorder items and blocks only using a special UI element for this.
5. Ensure that the correct data is sent to the API

## Expected Result

1. An admin can reorder items & blocks on the /about-us or similar page
2. Each block and list item has a drag handle icon 
3. Dragging via the handle reorders items; clicking/editing the block content is unaffected
4. The correct order is reflected in the API request payload on save

## Extra

The link for Figma implementation on the about-us page, where you can find all icons & understand the logic: 
https://www.figma.com/design/IDQ5XmNo6PQBmBlL3DBm6Q/Admin-Liatoshynsky-Foundation?node-id=2935-22375&t=w9lPgGSJUO9nSqIH-0

Issue #469, opened 2026-04-22
(SP: ) Implement file renaming functionality for assets in files page


dest411:
## Description

The user should be able to rename the selected file (asset) directly from the file details panel (FileInfoSidebar). To do this, you need to implement a modal window with a name editing form and connect an existing GraphQL mutation.

## Steps

1. Clicking on the pencil icon ("Rename") in the FileInfoSidebar opens the RenameFileModal modal window.
2. The modal window displays an input field, which is filled with the current file name by default.
3. When clicking on the "Cancel" button (or the cross), the modal window closes without saving the changes.
4. When clicking on the "Save" button, a request is sent to the backend to update the file name.
5. While the request is being executed, the "Save" button goes into the loading state (disabled / loader) to avoid double clicks.
6. After a successful response from the server, the modal window closes.
7. After a successful rename, a green snackbar/toaster appears with the text: "File successfully renamed".
8. The new file name is automatically updated in the UI (on the card and in the sidebar) thanks to the Apollo cache.

## Expected Result

The user can successfully change the name of the selected file via a modal window called from the sidebar. After saving, the new name is instantly updated throughout the interface (on cards and in the sidebar), and the user receives a toaster notification about the successful operation.

<img width="219" height="91" alt="Image" src="https://github.com/user-attachments/assets/79675446-20bc-4dd2-a25c-23f0b2a1242e" />

<img width="762" height="257" alt="Image" src="https://github.com/user-attachments/assets/9acc5db8-f094-4ccb-9e98-f4ba890d9a57" />

PR #730, opened 2026-07-24
fix: enforce meta title length limits (2-150 chars) on create / update news

LightOrden:
## Description

Closes https://github.com/Liatoshynsky-Foundation/lf-manual-tests/issues/469

Fixes lf-manual-tests#469: the GraphQL API (`createNews`/`updateNews`)
accepted a meta `title` (uk/en) longer than the documented 150-character
limit, silently saving it as the public H1.

This PR adds server-side and client-side max-length validation (150
characters) for the news title, additive to the min-length validation
already merged in #695 — no duplication of that logic.

## Changes

- `src/interfaces/graphql/resolvers/news/newsMutation.ts`: add
  `validateTitleMaxLength()`, called in both `createNews` and
  `updateNews`, checking `title.uk` and `title.en` independently.
  Throws `GraphQLError` with `extensions.code: 'BAD_USER_INPUT'`
  (matches the convention used elsewhere, e.g. `opusMutation.ts`).
- `src/interfaces/graphql/resolvers/news/newsMutation.ts`: add
  `trimLocalizedTitle()` — the title is trimmed before being persisted,
  so the stored value always matches what was validated (avoids a
  151-raw-character title with a trailing space being accepted as
  "150 characters" without normalizing storage).
- `src/constants/errors.ts`: add `TITLE_TOO_LONG_FOR_SLUG` message.
- `app/constants/publications.ts`: add `META_TITLE_MAX_LENGTH = 150`.
- `app/constants/errors.ts`: add `maxLength` message to `seoFormErrors`
  (uk/en).
- `SeoMetadataForm.tsx`: validate title length against the max on blur.
- `SeoBaseFields.tsx`: enforce `maxLength` on the meta title `TextField`
  via `slotProps.htmlInput`.
- Tests added/extended in `newsMutation.test.ts` and
  `SeoMetadataForm.test.tsx` covering: exactly 150 chars (valid), 151
  chars (rejected, both uk/en), and 150 chars + trailing whitespace
  (accepted and trimmed on store).

## Manual test steps

1. `createNews`/`updateNews` with `title.uk` or `title.en` at 151
   characters → expect `errors[0].extensions.code === "BAD_USER_INPUT"`.
2. Same with exactly 150 characters → expect success.
3. Same with 150 characters + trailing space → expect success, and the
   stored title has no trailing space.

## Related

- #602, #680 — drag-and-drop reordering (unrelated, referenced for
  context only)
- #695 — min-length validation this PR builds on top of (not duplicated)

## Video / Screenshots

image
image

### DIFF
--- a/src/interfaces/graphql/resolvers/news/newsMutation.ts
+++ b/src/interfaces/graphql/resolvers/news/newsMutation.ts
@@ -66,7 +66,9 @@ const validateTitleMaxLength = (title: LocalizedString | undefined): void => {
   (['uk', 'en'] as const).forEach((lang) => {
     const value = title[lang];
     if (typeof value === 'string' && value.trim().length > TITLE_MAX_LENGTH) {
-      throw new Error(newsServiceErrors.TITLE_TOO_LONG_FOR_SLUG);
+      throw new GraphQLError(newsServiceErrors.TITLE_TOO_LONG_FOR_SLUG, {
+        extensions: { code: 'BAD_USER_INPUT' }
+      });
     }
   });
 };
--- a/src/interfaces/graphql/resolvers/news/newsMutation.ts
+++ b/src/interfaces/graphql/resolvers/news/newsMutation.ts
@@ -58,6 +58,19 @@ const processContentFields = async (input: UpdateNewsGQLInput, updateData: Updat
   }
 };
 
+const TITLE_MAX_LENGTH = 150;
+
+const validateTitleMaxLength = (title: LocalizedString | undefined): void => {
+  if (!title) return;
+
+  (['uk', 'en'] as const).forEach((lang) => {
+    const value = title[lang];
+    if (typeof value === 'string' && value.trim().length > TITLE_MAX_LENGTH) {
+      throw new Error(newsServiceErrors.TITLE_TOO_LONG_FOR_SLUG);
+    }
+  });
+};
+
 const endpointHandler = endpointRepositoryHandler('newsRepository');
 
 export const NewsMutation = {
@@ -77,6 +90,8 @@ export const NewsMutation = {
     } else if (titleForSlug.trim().length < 2) {
       throw new Error(newsServiceErrors.TITLE_TOO_SHORT_FOR_SLUG);
     }
+
+    validateTitleMaxLength(input.title);
 
     const slug = await generateUniqueSlug(titleForSlug, {
       checkExists: async (slug: string) => {
@@ -142,6 +157,8 @@ export const NewsMutation = {
       } else if (titleForSlug.trim().length < 2) {
         throw new Error(newsServiceErrors.TITLE_TOO_SHORT_FOR_SLUG);
       }
+
+      validateTitleMaxLength(input.title);
 
       await processSlugUpdate(id, input.title, repo, updateData);
     }
--- a/src/interfaces/graphql/resolvers/news/newsMutation.ts
+++ b/src/interfaces/graphql/resolvers/news/newsMutation.ts
@@ -73,6 +73,24 @@ const validateTitleMaxLength = (title: LocalizedString | undefined): void => {
   });
 };
 
+// Trims uk/en title values so the persisted title always matches what was validated
+// (validation checks trimmed length; without this, a title with trailing/leading
+// whitespace could be stored with a raw length different from its validated length).
+const trimLocalizedTitle = (title: LocalizedString | undefined): LocalizedString | undefined => {
+  if (!title) return title;
+
+  const trimmed: LocalizedString = { ...title };
+
+  (['uk', 'en'] as const).forEach((lang) => {
+    const value = title[lang];
+    if (typeof value === 'string') {
+      trimmed[lang] = value.trim();
+    }
+  });
+
+  return trimmed;
+};
+
 const endpointHandler = endpointRepositoryHandler('newsRepository');
 
 export const NewsMutation = {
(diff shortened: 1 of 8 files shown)


src/interfaces/graphql/resolvers/news/newsMutation.ts
```diff
@@ -58,6 +58,19 @@ const processContentFields = async (input: UpdateNewsGQLInput, updateData: Updat
...
+  if (!title) return;
+
+  (['uk', 'en'] as const).forEach((lang) => {
+    const value = title[lang];
+    if (typeof value === 'string' && value.trim().length > TITLE_MAX_LENGTH) {
+      throw new Error(newsServiceErrors.TITLE_TOO_LONG_FOR_SLUG);
+    }
+  });
+};
+
```
dest411, line 75: The PR description states invalid input "throws a GraphQLError with BAD_USER_INPUT", and the manual test steps expect errors[0].extensions.code === "BAD_USER_INPUT". But this code throws a plain new Error(...), not a GraphQLError with that extension code set.

dest411 (changes requested):

src/interfaces/graphql/resolvers/news/newsMutation.ts
LightOrden: I corrected your comment in the last commit

dest411 (approved):
Good work!

src/interfaces/graphql/resolvers/news/newsMutation.ts
```diff
@@ -58,6 +58,21 @@ const processContentFields = async (input: UpdateNewsGQLInput, updateData: Updat
...
 };
 
+const TITLE_MAX_LENGTH = 150;
+
+const validateTitleMaxLength = (title: LocalizedString | undefined): void => {
+  if (!title) return;
+
+  (['uk', 'en'] as const).forEach((lang) => {
+    const value = title[lang];
+    if (typeof value === 'string' && value.trim().length > TITLE_MAX_LENGTH) {
```
Halyna-Trush, line 68: The frontend and backend calculate the title length differently. `maxLength={150}` counts all characters, while the API checks the length after applying `trim()`. For example, a string containing 150 letters followed by a trailing space has an actual length of 151, but the API considers it valid.
Perhaps we should check `value.length` to keep the validation consistent with the frontend, or trim the spaces before validation and store the trimmed value. I also think it would be worth adding a test for `'a'.repeat(150) + ' '`.

Halyna-Trush:
The PR description says that `validateLocalizedTitleLength` was added to `src/interfaces/graphql/resolvers/helpers.ts`. However, this file is not included in the PR, and I cannot find `validateLocalizedTitleLength` in the changed files.

The implementation currently adds only the maximum-length validation in `newsMutation.ts`. Could you please check whether the helper and the minimum-length validation for both `title.uk` and `title.en` are missing from the PR?

LightOrden:
> The PR description says that `validateLocalizedTitleLength` was added to `src/interfaces/graphql/resolvers/helpers.ts`. However, this file is not included in the PR, and I cannot find `validateLocalizedTitleLength` in the changed files.
> 
> The implementation currently adds only the maximum-length validation in `newsMutation.ts`. Could you please check whether the helper and the minimum-length validation for both `title.uk` and `title.en` are missing from the PR?

Good catch — the PR description is outdated. It referred to an earlier
implementation attempt (a shared `validateLocalizedTitleLength` helper
in `helpers.ts`, checking both min and max length) that was scrapped
when this branch was reset onto `develop` after #695 was merged.

The current implementation validates max length inline in
`newsMutation.ts` (`validateTitleMaxLength`), covering both `title.uk`
and `title.en` — see the corresponding tests in `newsMutation.test.ts`
for both languages.

Regarding min-length: that check is pre-existing code from #695
(already merged to `develop`, not part of this diff). You're right
that it only validates `title.uk` via `extractTitleForSlug` and never
checks `title.en`'s length independently — but that's out of scope
for this PR, which targets the max-length bug (#469) specifically.
I'll open a follow-up issue to track the `.en` min-length gap rather
than expanding this PR's scope.

I've updated the PR description to match the current diff.

src/interfaces/graphql/resolvers/news/newsMutation.ts
LightOrden: Fixed: the title is now trimmed before being
validated/stored, so a 150-character title with a trailing space is
accepted and persisted as exactly 150 trimmed characters — no
mismatch between the raw and trimmed length anymore. Added tests for
'a'.repeat(150) + ' ' covering both createNews and updateNews.

Halyna-Trush:
> Good catch — the PR description is outdated. It referred to an earlier implementation attempt (a shared `validateLocalizedTitleLength` helper in `helpers.ts`, checking both min and max length) that was scrapped when this branch was reset onto `develop` after #695 was merged.
> 
> The current implementation validates max length inline in `newsMutation.ts` (`validateTitleMaxLength`), covering both `title.uk` and `title.en` — see the corresponding tests in `newsMutation.test.ts` for both languages.
> 
> Regarding min-length: that check is pre-existing code from #695 (already merged to `develop`, not part of this diff). You're right that it only validates `title.uk` via `extractTitleForSlug` and never checks `title.en`'s length independently — but that's out of scope for this PR, which targets the max-length bug (#469) specifically. I'll open a follow-up issue to track the `.en` min-length gap rather than expanding this PR's scope.
> 
> I've updated the PR description to match the current diff.

Thanks for the clarification!

dest411 (approved):

annak413 (approved):

PR #754, opened 2026-07-29
Feature: Add standalone composition creation from the Creativity page

### LINKED ISSUE #727: (SP: 2) Allow creating standalone compositions from the /creativity page
## Description

Currently, compositions can only be created while creating or editing an opus group. Users should also be able to create standalone compositions directly from the **"Створити"** button on the **/creativity** page.

Implement a modal using the existing `CompositionModal` component in `mode="create"` and add the required backend/frontend logic to support composition creation outside of the opus creation flow.

## Expected Result

- Users can create a standalone composition directly from the **/creativity** page.
- The composition is created successfully without requiring an opus group.
- After successful creation, the new composition appears in the compositions table.

## Extra

Image

vladashvch:
## Description

Implements standalone composition creation from the **/creativity** page without requiring an opus group.

### Frontend
- Added the ability to create a standalone composition from the **"Створити"** button on the **/creativity** page.
- Reused the existing `CompositionModal` component in `mode="create"`.
- Added the required frontend flow for creating compositions outside of the opus creation workflow.
- After a successful creation, the compositions table is updated to display the newly created composition.

### Backend
- Added dedicated composition creation mutation endpoint for standalone composition creation.

### Data model
- Updated the composition media structure:
  - **SheetMusic**
    - `url` is no longer required.
    - A SheetMusic can contain:
      - `name`
      - `url`
      - `name + date`
      - `url + date`
      - `name + url`
      - `name + url + date`
  - **Audio**
    - Both `name` and `url` are now required.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other:

## How to test

1. Check out this branch.
2. Navigate to the `/creativity` page.
3. Click the **"Створити"** button.
4. Select **Composition**.
5. Fill in the required fields and create a composition.
6. Verify that the composition is created successfully.
7. Verify that the new composition appears in the compositions table.
8. Verify the updated media validation:
   - Notes can be saved without a URL.
   - Audio requires both a name and a URL.

### How it works now

https://github.com/user-attachments/assets/6435ee46-1d91-4fec-b17e-49118c7fe242

image

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues

---

Closes #727


Halyna-Trush:
Good job! Please fix the Sonar issue.

app/shared/components/forms/opus-details-block/composition-modal/CompositionModal.tsx
```diff
@@ -151,7 +202,7 @@ export default function CompositionModal({
           InputProps={{
             endAdornment: composition.name ? (
               <InputAdornment position="end">
-                <IconButton aria-label="Очистити" size="small" onClick={() => updateField('name', '')}>
+                <IconButton size="small" onClick={() => updateField('name', '')}>
```
Halyna-Trush: I think we should restore the `aria-label` attributes for icon-only buttons, including clearing the title, deleting audio, uploading and deleting sheet music, and removing an attached file. Without accessible names, a screen reader announces each of them only as “button,” so users cannot understand what each action does. The corresponding text has already been added to `COMPOSITION_MODAL_TEXTS`, so it can be used for the `aria-label` values.

app/(logged_in)/creativity/(composition)/useCreateWorkAction.ts
```diff
@@ -0,0 +1,103 @@
...
+      ),
+    [mutate]
+  );
+
+  return [createComposition, meta] as const;
+};
+
+const mapToCreateInput = (work: OpusCompositionData): CreateCompositionInput => ({
+  name: { uk: work.name, en: work.name },
+  year: work.year ? Number(work.year) : undefined,
```
Halyna-Trush: It looks like the year field currently accepts arbitrary text, while the submitted value is converted using `Number()`. For example, entering `abc` produces `NaN`, which cannot be sent as a GraphQL `Int`, so the creation request fails with a generic error. Could we please restrict this field to valid integers or validate it before submission and display an appropriate field-level error?

src/interfaces/graphql/resolvers/compositions/compositionsMutation.ts
```diff
@@ -0,0 +1,54 @@
...
+    } = context.requestContainer.cradle;
+
+    const compositionData = mapCompositionInput(input);
+
+    const composition = await repo.create(compositionData);
+    if (!composition) {
+      throw new Error(compositionsServiceErrors.COMPOSITION_NOT_CREATED);
+    }
+
+    await opusRepo.moveCompositionsToCompositionsOpus([composition.id]);
```
Halyna-Trush, line 50: It seems possible to create an orphaned record here. If the composition is saved successfully but `moveCompositionsToCompositionsOpus` fails, the mutation will return an error while the newly created composition remains in the database and does not appear in the table. Retrying the operation may then create a duplicate. Could we run these operations in a single transaction or remove the created composition if linking it to the corresponding opus fails?

dest411 (approved):

app/shared/components/forms/opus-details-block/composition-modal/CompositionModal.tsx
vladashvch: Yea, good point! Fixed it

app/(logged_in)/creativity/(composition)/useCreateWorkAction.ts
vladashvch: Thanks for your catch! Fixed it

src/interfaces/graphql/resolvers/compositions/compositionsMutation.ts
vladashvch: Yes, I agree that wrapping these operations in a transaction (or implementing a rollback) would make the flow more robust. However, I think it's better to handle transaction support in a separate task. The goal of this task is to enable standalone composition creation, while introducing transactions requires broader infrastructure changes

app/(logged_in)/creativity/(composition)/useCreateWorkAction.ts
Halyna-Trush: `NaN` is no longer submitted, but the field still accepts arbitrary text. If a user enters `abc`, the composition will be created without a year, and no validation error will be displayed. Am I correct in understanding that this is the expected behavior?

Halyna-Trush (approved):

Issue #501, opened 2026-05-01
(SP: 0.5) [Admin panel] Scrolling background is able when modal is active.


sofiatymchak:
**Environment:** macOs 15.7.5 (24G624)
Safari  18.6 (20621.3.11.11.3)
Chrome 147.0.7727.102
**Reproducible:** always.

**Steps to reproduce**

1. Go to https://lf-admin-stage-dzfpbxeuauh7fad2.polandcentral-01.azurewebsites.net/.
2. Login into admin.
3. Press “Новини та події” in navigation bar.
4. Press to “Create” dropdown and choose a “News” item.
5. Click “Change image” button.
6. 6.Modal is opened.
7. Scroll behind the modal.

**Actual result**
Scrolling is able.

**Expected result**
Scrolling is disabled when Modal is active.

Jam: https://jam.dev/c/a393cc49-c445-4072-b73f-9522f5f9c59b.

**User story and test case links**
E.g.:US-Admin-IMG-1 — Open the “Select Image” modal from a page block #1048
https://github.com/Liatoshynsky-Foundation/lf-client/issues/1048

**Labels to be added**
"Bug", 
Priority: Medium
Severity: Low

sofiatymchak:
Retested and passed by QA.

PR #763, opened 2026-08-04
Fix news meta description validation

exity2905:
## Description
This PR fixes bug #501 by adding backend validation for News meta description length.

News can no longer be created when `description.uk` or `description.en` contains fewer than 2 characters or more than 250 characters. The API returns `BAD_USER_INPUT` with invalid description fields.

Closes [Liatoshynsky-Foundation/lf-manual-tests#501](https://github.com/Liatoshynsky-Foundation/lf-manual-tests/issues/501)

## How to test
1. Run `npm test -- newsMutation.test.ts --coverage=false`.
2. In Postman, send `CreateNews` with `description.uk: "T"` and `description.en: "T"`.
3. Verify the API returns `BAD_USER_INPUT` and `fields: ["description.uk", "description.en"]`.

## Video / Screenshots
Not applicable. API validation change.

## Checklist
- [x] Tests pass successfully
- [x] No secret keys of our application
- [x] No Sonar errors

### DIFF
--- a/src/interfaces/graphql/resolvers/news/newsMutation.test.ts
+++ b/src/interfaces/graphql/resolvers/news/newsMutation.test.ts
@@ -163,6 +163,25 @@ describe('NewsMutation Resolvers', () => {
     it('should throw GraphQLError for updateNews if user is unauthenticated', async () => {
       await expect(NewsMutation.updateNews({}, { id: '1', input: {} }, userContext)).rejects.toThrow();
     });
+
+    it('should throw GraphQLError with BAD_USER_INPUT if description has fewer than 2 characters (bug #501)', async () => {
+      expect.assertions(5);
+
+      const invalidInput = {
+        ...baseInput,
+        description: { uk: 'T', en: 'T' }
+      };
+
+      try {
+        await NewsMutation.createNews({}, { input: invalidInput }, adminContext);
+      } catch (error) {
+        expect(error).toBeInstanceOf(GraphQLError);
+        expect((error as GraphQLError).message).toBe(newsServiceErrors.DESCRIPTION_LENGTH_INVALID);
+        expect((error as GraphQLError).extensions.code).toBe('BAD_USER_INPUT');
+        expect((error as GraphQLError).extensions.fields).toEqual(['description.uk', 'description.en']);
+        expect(mockRepo.create).not.toHaveBeenCalled();
+      }
+    });
   });
 
   describe('createNews', () => {
--- a/src/interfaces/graphql/resolvers/news/newsMutation.test.ts
+++ b/src/interfaces/graphql/resolvers/news/newsMutation.test.ts
@@ -164,40 +164,32 @@ describe('NewsMutation Resolvers', () => {
       await expect(NewsMutation.updateNews({}, { id: '1', input: {} }, userContext)).rejects.toThrow();
     });
 
-    it('should throw GraphQLError with BAD_USER_INPUT if description has fewer than 2 characters', async () => {
-
+    it.each([
+      {
+        caseName: 'has fewer than 2 characters',
+        description: { uk: 'T', en: 'T' },
+        fields: ['description.uk', 'description.en']
+      },
+      {
+        caseName: 'exceeds 250 characters',
+        description: { uk: 'a'.repeat(251), en: 'Valid description' },
+        fields: ['description.uk']
+      }
+    ])('should throw GraphQLError with BAD_USER_INPUT if description $caseName', async ({ description, fields }) => {
       const invalidInput = {
         ...baseInput,
-        description: { uk: 'T', en: 'T' }
+        description
       };
 
-      try {
-        await NewsMutation.createNews({}, { input: invalidInput }, adminContext);
-        throw new Error('Expected createNews to throw');
-      } catch (error) {
-        expect(error).toBeInstanceOf(GraphQLError);
-        expect((error as GraphQLError).message).toBe(newsServiceErrors.DESCRIPTION_LENGTH_INVALID);
-        expect((error as GraphQLError).extensions.code).toBe('BAD_USER_INPUT');
-        expect((error as GraphQLError).extensions.fields).toEqual(['description.uk', 'description.en']);
-        expect(mockRepo.create).not.toHaveBeenCalled();
-      }
-    });
-    it('should throw GraphQLError with BAD_USER_INPUT if description exceeds 250 characters', async () => {
-      const invalidInput = {
-        ...baseInput,
-        description: { uk: 'a'.repeat(251), en: 'Valid description' }
-      };
+      await expect(NewsMutation.createNews({}, { input: invalidInput }, adminContext)).rejects.toMatchObject({
+        message: newsServiceErrors.DESCRIPTION_LENGTH_INVALID,
+        extensions: {
+          code: 'BAD_USER_INPUT',
+          fields
+        }
+      });
 
-      try {
-        await NewsMutation.createNews({}, { input: invalidInput }, adminContext);
-        throw new Error('Expected createNews to throw');
-      } catch (error) {
-        expect(error).toBeInstanceOf(GraphQLError);
-        expect((error as GraphQLError).message).toBe(newsServiceErrors.DESCRIPTION_LENGTH_INVALID);
-        expect((error as GraphQLError).extensions.code).toBe('BAD_USER_INPUT');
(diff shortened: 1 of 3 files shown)


src/interfaces/graphql/resolvers/news/newsMutation.ts
```diff
@@ -60,6 +60,31 @@ const processContentFields = async (input: UpdateNewsGQLInput, updateData: Updat
...
+  if (!description) return;
+
+  const invalidFields = (['uk', 'en'] as const)
+    .filter((lang) => {
+      const value = description[lang];
+      const length = typeof value === 'string' ? value.trim().length : 0;
+
+      return length < DESCRIPTION_MIN_LENGTH || length > DESCRIPTION_MAX_LENGTH;
+    })
+    .map((lang) => `description.${lang}`);
```
dest411: When performing partial updates via updateNews with only one locale provided (e.g. { uk: "Updated description text" }), description.en is undefined.
Currently, const length = typeof value === 'string' ? value.trim().length : 0; evaluates undefined as length = 0, which satisfies 0 < DESCRIPTION_MIN_LENGTH. This causes updateNews to incorrectly reject valid partial description updates with a BAD_USER_INPUT error for description.en.

dest411 (changes requested):

src/interfaces/graphql/resolvers/news/newsMutation.test.ts
```diff
@@ -163,6 +163,25 @@ describe('NewsMutation Resolvers', () => {
     it('should throw GraphQLError for updateNews if user is unauthenticated', async () => {
       await expect(NewsMutation.updateNews({}, { id: '1', input: {} }, userContext)).rejects.toThrow();
     });
+
+    it('should throw GraphQLError with BAD_USER_INPUT if description has fewer than 2 characters (bug #501)', async () => {
```
telare: we can omit clarification of a bug that is related here)

src/interfaces/graphql/resolvers/news/newsMutation.test.ts
```diff
@@ -163,6 +163,25 @@ describe('NewsMutation Resolvers', () => {
...
+      expect.assertions(5);
```
telare: Can u clarify the necessity of using that here?

telare:
1) We should not use `Projects` in a PR
2) It is better to request a review from only your group and other active developers like Denis, Vadym, etc
3) To correctly close the issue related to the if-manual-tests repo, you should write `Closes [link to your issue in if-manual-tests`.
Like it has been done here: https://github.com/Liatoshynsky-Foundation/lf-client/pull/1543

telare (changes requested):

src/interfaces/graphql/resolvers/news/newsMutation.test.ts
telare: Also, you can add a test that checks for an error on more than 250 characters. Moreover, tests for partial success for a locale (uk with a valid length, while en is invalid or vice versa) will be great)

telare (changes requested):

src/interfaces/graphql/resolvers/news/newsMutation.ts
exity2905: Addressed in the latest commit. Description validation now skips locales that are not provided, so partial updateNews requests with only one locale do not fail because the other locale is undefined.

I also added a test for updateNews with only one provided valid locale.

src/interfaces/graphql/resolvers/news/newsMutation.test.ts
exity2905: Fixed, removed the issue reference from the test name.

src/interfaces/graphql/resolvers/news/newsMutation.test.ts
exity2905: Removed expect.assertions and added an explicit throw after createNews, so the test still fails if createNews does not throw.

src/interfaces/graphql/resolvers/news/newsMutation.test.ts
exity2905: Added a test for description longer than 250 characters, and added a test for updateNews with only one provided valid locale.

dest411 (approved):

LightOrden (changes requested):
Thanks for the PR, the logic looks solid overall, but please address these before merging:

1. **Duplicated validation logic.** `validateDescriptionLength` and `validateTitleMaxLength` do essentially the same thing but behave differently: title throws on the first invalid field, description collects all invalid fields first. Please align both to the same behavior (collect-all is preferable) and extract a shared function to avoid duplicating the logic.

2. **Missing trim before saving `description`.** `title` has `trimLocalizedTitle` applied before being written to the DB, but `description` doesn't. Validation checks length on `.trim()`, yet the raw value (with whitespace) gets saved. Please trim `description` before saving, same as `title`.

src/interfaces/graphql/resolvers/news/newsMutation.test.ts
```diff
@@ -163,6 +163,56 @@ describe('NewsMutation Resolvers', () => {
...
+
+      try {
+        await NewsMutation.createNews({}, { input: invalidInput }, adminContext);
+        throw new Error('Expected createNews to throw');
+      } catch (error) {
+        expect(error).toBeInstanceOf(GraphQLError);
+        expect((error as GraphQLError).message).toBe(newsServiceErrors.DESCRIPTION_LENGTH_INVALID);
+        expect((error as GraphQLError).extensions.code).toBe('BAD_USER_INPUT');
+        expect((error as GraphQLError).extensions.fields).toEqual(['description.uk']);
+        expect(mockRepo.create).not.toHaveBeenCalled();
```
telare: You mostly repeat this logic in 2 tests.
U can leverage `it.each()` to avoid code duplication.

src/interfaces/graphql/resolvers/news/newsMutation.test.ts
```diff
@@ -163,6 +163,56 @@ describe('NewsMutation Resolvers', () => {
...
+    });
+    it('should throw GraphQLError with BAD_USER_INPUT if description exceeds 250 characters', async () => {
+      const invalidInput = {
+        ...baseInput,
+        description: { uk: 'a'.repeat(251), en: 'Valid description' }
+      };
+
+      try {
+        await NewsMutation.createNews({}, { input: invalidInput }, adminContext);
+        throw new Error('Expected createNews to throw');
```
telare: And u can leverage built-in Jest `.rejects.toMatchObject` like this: 
`await expect(NewsMutation.createNews({}, { input: invalidInput }, adminContext)).rejects.toMatchObject`. This removes the `try catch` block

src/interfaces/graphql/resolvers/news/newsMutation.ts
```diff
@@ -60,34 +60,87 @@ const processContentFields = async (input: UpdateNewsGQLInput, updateData: Updat
...
+  maxLength?: number;
+};
+
+const getInvalidLocalizedLengthFields = (
+  value: LocalizedString | undefined,
+  { fieldName, minLength, maxLength }: LocalizedLengthValidationOptions
+): string[] => {
+  if (!value) return [];
+
+  return (['uk', 'en'] as const)
```
telare: Minor, but to improve readability, u can make an additional variable to store the result of the array mutations (filter and map

telare (commented):
Good work with my old comments! But, I've found new places to improve your changes)

LightOrden (approved):
LGFM

src/interfaces/graphql/resolvers/news/newsMutation.test.ts
exity2905: Updated, combined the duplicated description validation cases with it.each().

src/interfaces/graphql/resolvers/news/newsMutation.test.ts
exity2905: Updated, replaced the manual try/catch with Jest rejects.toMatchObject().

src/interfaces/graphql/resolvers/news/newsMutation.ts
exity2905: Updated, extracted the filtered fields into invalidFields for readability.

telare (approved):
good first PR! Excellent work with comments

PR #765, opened 2026-08-05
feature: implement actions menu functionality for compositions inside opus

### LINKED ISSUE #729: (SP: 1) Implement actions for opus composition menu
## Description

The actions menu for compositions inside an opus is already displayed, but its actions are not implemented.

Implement the existing menu functionality to support:
- Share composition
- Edit composition
- Delete composition

Use the existing `CompositionModal` for editing.

## Expected Result

- Users can share a composition from an opus.
- Users can edit a composition from an opus.
- Users can delete a composition from an opus after confirmation.

## Extra

Image

vladashvch:
## Description

Implements functionality for the composition actions menu inside opus groups. Users can now share, edit, and unlink compositions directly from an opus group on the table using the existing modals and sharing utilities.

### What was changed

- Connected the composition actions menu (`WorkMenuItems`) inside opus groups to functional handlers.
- Integrated the existing `CompositionModal` in edit mode for compositions within an opus.
- Added support for sharing individual compositions via generated URL parameters.
- Implemented the `unlinkComposition` GraphQL mutation, backend resolver, and frontend hook to remove a composition from an opus group without deleting the composition itself.
- Added a dedicated confirmation modal (`DeleteCardModal`) for unlinking compositions from an opus group.
- Added success/error feedback via toasts for all supported actions.
- Updated the `CompositionModal` year input field to restrict user entry exclusively to numerical digits using a regular expression check.

## Type of change

- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Check out this branch.
2. Navigate to the `/creativity` page.
3. Expand an opus group.
4. Open the actions menu (three dots) for any composition inside the group.
5. Verify the following actions:
   - **Share** – confirm that a shareable link containing the composition id value is generated/copied.
   - **Edit** – confirm that the `CompositionModal` opens with the composition data pre-filled and that changes are saved successfully.
   - **Delete (Unlink)** – confirm that the confirmation modal appears and that, after confirmation, the composition is unlinked from the opus group and becomes an independent composition again.

## Video / Screenshots

https://github.com/user-attachments/assets/3a73ed2a-5401-44aa-b628-0cd4ec3902fc

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues

---

Closes #729

### DIFF
--- a/app/shared/components/forms/opus-details-block/composition-modal/CompositionModal.tsx
+++ b/app/shared/components/forms/opus-details-block/composition-modal/CompositionModal.tsx
@@ -225,7 +225,12 @@ export default function CompositionModal({
           <TextField
             label={COMPOSITION_MODAL_LABELS.year}
             value={composition.year}
-            onChange={(event) => updateField('year', event.target.value)}
+            onChange={(event) => {
+              const val = event.target.value;
+              if (/^\d*$/.test(val)) {
+                updateField('year', val);
+              }
+            }}
             size="small"
             sx={styles.field}
           />
--- a/src/interfaces/graphql/resolvers/opus/opusMutation.test.ts
+++ b/src/interfaces/graphql/resolvers/opus/opusMutation.test.ts
@@ -851,7 +851,7 @@ describe('OpusMutation Resolvers', () => {
       await expect(
         OpusMutation.unlinkComposition({}, { opusId: OPUS_ID, compositionId: COMPOSITION_ID_1 }, adminContext)
       ).rejects.toThrow(
-        new GraphQLError(opusServiceErrors.COMPOSITION_NOT_FOUND_IN_OPUS ?? 'Composition not found in this opus', {
+        new GraphQLError(opusServiceErrors.COMPOSITION_NOT_FOUND_IN_OPUS, {
           extensions: { code: 'BAD_USER_INPUT' }
         })
       );
@@ -888,5 +888,29 @@ describe('OpusMutation Resolvers', () => {
       expect(mockedOrderCompositionsByIds).toHaveBeenCalledWith([COMPOSITION_ID_2], [MOCK_COMPOSITION_2]);
       expect(result).toEqual({ ...updatedOpusEntity, compositions: [MOCK_COMPOSITION_2] });
     });
+
+    it.each([undefined, null])(
+      'should handle %s opus.compositions gracefully and throw error when composition is not found',
+      async (compositionsValue) => {
+        const opusWithoutCompositions = {
+          ...MOCK_OPUS_ENTITY,
+          compositions: compositionsValue as unknown as string[]
+        };
+
+        mockOpusRepo.findById.mockResolvedValue(opusWithoutCompositions as unknown as Opus);
+
+        await expect(
+          OpusMutation.unlinkComposition(
+            {},
+            { opusId: OPUS_ID, compositionId: COMPOSITION_ID_1 },
+            adminContext
+          )
+        ).rejects.toThrow(
+          new GraphQLError(opusServiceErrors.COMPOSITION_NOT_FOUND_IN_OPUS, {
+            extensions: { code: 'BAD_USER_INPUT' }
+          })
+        );
+      }
+    );
   });
 });
--- a/src/interfaces/graphql/resolvers/opus/opusMutation.test.ts
+++ b/src/interfaces/graphql/resolvers/opus/opusMutation.test.ts
@@ -817,4 +817,76 @@ describe('OpusMutation Resolvers', () => {
       expect(result).toBe(true);
     });
   });
+
+  describe('unlinkComposition', () => {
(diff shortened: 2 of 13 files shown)


src/interfaces/graphql/resolvers/opus/opusMutation.ts
```diff
@@ -467,5 +468,37 @@ export const OpusMutation = {
...
+    assertAuthenticated(context);
+
+    const {
+      opusRepository: repo,
+      compositionsRepository: compositionsRepo
+    } = context.requestContainer.cradle;
+
+    const existingOpus = await findExistingOpus(repo, opusId);
+
+    const currentCompositions = (existingOpus.compositions ?? []).map((id) => id.toString());
```
telare: I think that this could lead to runtime errors. Possibly, if compositions contain null/undefined values, you will get an error `TypeError: Cannot read properties of undefined (reading 'toString')`. So, my suggestion is to filter the compositions before calling `toString()`. It will ensure no runtime errors

app/shared/components/forms/opus-details-block/composition-modal/CompositionModal.tsx
```diff
@@ -225,7 +225,12 @@ export default function CompositionModal({
           <TextField
             label={COMPOSITION_MODAL_LABELS.year}
             value={composition.year}
-            onChange={(event) => updateField('year', event.target.value)}
+            onChange={(event) => {
+              const val = event.target.value;
+              if (/^\d*$/.test(val)) {
```
telare, line 230: I think u should cover this with tests, too, like `restricts non-numeric input in the year field`.

telare (changes requested):

src/interfaces/graphql/resolvers/opus/opusMutation.ts
DenisGordProgrammer: Agreed, this is a worthwhile defensive fix: a malformed or legacy opus reference should not make unlinking fail before the mutation can validate the requested composition. Filtering or otherwise guarding nullish IDs before `toString()` would make this path safer.

DenisGordProgrammer (approved):
Approved. The composition actions are connected end to end, unlinking preserves the composition while removing the opus relation, and the added tests cover the new handlers and modal flow. I also replied to the existing null-reference thread with a small defensive follow-up.

src/interfaces/graphql/resolvers/opus/opusMutation.test.ts
```diff
@@ -817,4 +817,76 @@ describe('OpusMutation Resolvers', () => {
...
+        ...MOCK_OPUS_ENTITY,
+        compositions: [COMPOSITION_ID_2]
+      };
+      mockOpusRepo.findById.mockResolvedValue(opusWithoutComp as unknown as Opus);
+
+      await expect(
+        OpusMutation.unlinkComposition({}, { opusId: OPUS_ID, compositionId: COMPOSITION_ID_1 }, adminContext)
+      ).rejects.toThrow(
+        new GraphQLError(opusServiceErrors.COMPOSITION_NOT_FOUND_IN_OPUS ?? 'Composition not found in this opus', {
+          extensions: { code: 'BAD_USER_INPUT' }
```
LightOrden, line 855: opusServiceErrors.COMPOSITION_NOT_FOUND_IN_OPUS is already a constantly defined string in src/constants/errors.ts, so the ?? 'Composition not found in this opus' fallback is dead code — it will never execute. Suggest removing it:
```
throw new GraphQLError(opusServiceErrors.COMPOSITION_NOT_FOUND_IN_OPUS, {
  extensions: { code: 'BAD_USER_INPUT' }
});
```
The same duplication also exists in the test (opusMutation.test.ts) — worth cleaning up there too.

LightOrden (changes requested):

src/interfaces/graphql/resolvers/opus/opusMutation.ts
vladashvch: Thanks for all your precise reviews! I've added the filter as you mentioned. That's a nice catch!

app/shared/components/forms/opus-details-block/composition-modal/CompositionModal.tsx
vladashvch: Thanks for pointing that out! I've added unit tests to cover non-numeric values for the year field as suggested

src/interfaces/graphql/resolvers/opus/opusMutation.test.ts
vladashvch: Good advice. Removed the redundant fallback in both the implementation and the test file. Thanks a lot!

LightOrden (approved):
LGFM

telare (approved):

PR #580, opened 2026-06-03
fix(ui): close content card dropdown on scroll

pALINchuk:
## Description

Fixed the content card dropdown behavior during page scroll.

Previously, when the user opened the "More options" dropdown on a content card and scrolled the page, the dropdown stayed fixed in the viewport and detached from the card. Now the dropdown automatically closes when scrolling starts.

Closes [#565](https://github.com/Liatoshynsky-Foundation/lf-admin/issues/565)

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Navigate to the Publications page.
2. Click the "More options" icon on any content card.
3. Scroll the page up or down using the mouse wheel or trackpad.
4. Verify that the dropdown menu closes as soon as scrolling starts.

## Checklist

- [x] Code compiles and runs correctly
- [x] Tests pass successfully
- [x] Linting checks are clean
- [ ] No Sonar issues
- [ ] Documentation updated (if applicable)
- [ ] All comments and requested changes addressed
- [x] Branch is up to date with the target branch
- [x] Linked issue/task included
- [ ] Task moved to the review column on the board

---

PR #783, opened 2026-08-11
fix: add validation for title on createNews/updateNews

MargoMatskevych:
## Description

Fixes validation for the English SEO title in the Create News mutation.

Previously, a title with only 1 character could be accepted, which could lead to an invalid publication being created. The validation now correctly handles the minimum title length.

## Type of change

- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring / Tech debt
- [ ] CI/CD improvement
- [ ] Other: 

## How to test

1. Execute the Login request in Postman.
2. Verify that authorization data is successfully saved in Postman variables.
3. Open the GraphQL request Mutations/Create News.
4. In GraphQL Variables, set title.en to 1 character while keeping all other required fields valid:
```
{
  "input": {
    "adminTitle": "Test Yana",
    "title": {
      "uk": "Це тестовий заголовок meta новини українською мовою",
      "en": "A"
    }
  }
}
```
5. Send the request.
6. Check the API response.
7. Open the created publication on the website/client side.

## Video / Screenshots
Screenshot 2026-08-11 at 18 56 35

## Checklist

- [x] Tests pass successfully
- [x] No Sonar issues

 Closes: [#580](https://github.com/Liatoshynsky-Foundation/lf-manual-tests/issues/580)
---

### DIFF
--- a/src/interfaces/graphql/resolvers/news/newsMutation.ts
+++ b/src/interfaces/graphql/resolvers/news/newsMutation.ts
@@ -58,6 +58,7 @@ const processContentFields = async (input: UpdateNewsGQLInput, updateData: Updat
   }
 };
 
+const TITLE_MIN_LENGTH = 2;
 const TITLE_MAX_LENGTH = 150;
 
 const DESCRIPTION_MIN_LENGTH = 2;
@@ -113,13 +114,14 @@ const validateDescriptionLength = (description: LocalizedString | undefined): vo
   throwBadUserInput(newsServiceErrors.DESCRIPTION_LENGTH_INVALID, invalidFields);
 };
 
-const validateTitleMaxLength = (title: LocalizedString | undefined): void => {
+const validateTitleLength = (title: LocalizedString | undefined): void => {
   const invalidFields = getInvalidLocalizedLengthFields(title, {
     fieldName: 'title',
-    maxLength: TITLE_MAX_LENGTH
+    maxLength: TITLE_MAX_LENGTH,
+    minLength: TITLE_MIN_LENGTH
   });
 
-  throwBadUserInput(newsServiceErrors.TITLE_TOO_LONG_FOR_SLUG, invalidFields);
+  throwBadUserInput(newsServiceErrors.TITLE_LENGTH_INVALID, invalidFields);
 };
 
 const trimLocalizedString = (value: LocalizedString | undefined): LocalizedString | undefined => {
@@ -160,11 +162,9 @@ export const NewsMutation = {
 
     if (!titleForSlug) {
       throw new Error(newsServiceErrors.TITLE_REQUIRED_FOR_SLUG);
-    } else if (titleForSlug.trim().length < 2) {
-      throw new Error(newsServiceErrors.TITLE_TOO_SHORT_FOR_SLUG);
     }
 
-    validateTitleMaxLength(trimmedInput.title);
+    validateTitleLength(trimmedInput.title);
     validateDescriptionLength(trimmedInput.description);
 
     const slug = await generateUniqueSlug(titleForSlug, {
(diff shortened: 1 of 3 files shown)


Yushchyk-Roman:
Link the PR to the Issue, Use `Closes #` and number of Issue

src/interfaces/graphql/resolvers/news/newsMutation.ts
```diff
@@ -238,11 +238,9 @@ export const NewsMutation = {
       const titleForSlug = extractTitleForSlug(trimmedInput.title);
       if (!titleForSlug) {
         throw new Error(newsServiceErrors.TITLE_REQUIRED_FOR_SLUG);
-      } else if (titleForSlug.trim().length < 2) {
-        throw new Error(newsServiceErrors.TITLE_TOO_SHORT_FOR_SLUG);
       }
 
-      validateTitleMaxLength(trimmedInput.title);
+      validateTitleLength(trimmedInput.title);
```
dest411: We should run` validateTitleLength(trimmedInput.title)` before `extractTitleForSlug`.
If a user sends empty or whitespace-only titles (e.g.` { uk: "", en: "" }`), `extractTitleForSlug `will return `null `first and throw a generic raw `Error(TITLE_REQUIRED_FOR_SLUG)`, this prevents GraphQL from returning a proper `BAD_USER_INPUT `error code containing the affected `fields`

dest411 (changes requested):

telare (approved):
congrats w the 1st PR, keep going!

src/interfaces/graphql/resolvers/news/newsMutation.ts
MargoMatskevych: 1. Extract `extractTitleForSlug` should return <empty string> (based on tests) 
Screenshot 2026-08-12 at 14 49 03

2. If we put it before `extractTitleForSlug` we will have no chance to receive `TITLE_REQUIRED_FOR_SLUG` error message. I mean it's not clear if it's required field or field with incorrect length of string

dest411: The main point here is error formatting, `TITLE_REQUIRED_FOR_SLUG` is a generic `new Error()`, so if it fires first, the UI just gets a raw server error without any field details to highlight

since `validateTitleLength` checks `minLength = 2`, an empty string (`""` with length 0) fails validation anyway and returns a proper `BAD_USER_INPUT` error with `fields: ['title.uk']`

running validation first simply guarantees the frontend always gets clean, field-specific errors

dest411:
@MargoMatskevych 
Here is the proof from Postman:

1. When `title.en` is **1 char** (`"T"`), the server correctly returns `BAD_USER_INPUT` with `fields: ["title.en"]`
image


2. When `title` is **empty** (`""`), because `extractTitleForSlug` runs first, it crashes with `INTERNAL_SERVER_ERROR` ("Title is required to generate a slug") instead of returning a proper `BAD_USER_INPUT` validation error
image


moving `validateTitleLength` before `extractTitleForSlug` fixes this and ensures consistent `BAD_USER_INPUT` responses for all invalid inputs

MargoMatskevych:
@dest411 , okay let's try you suggestion 
Screenshot 2026-08-12 at 17 48 05
and now let's test it 
1.  title with one char
Screenshot 2026-08-12 at 17 48 56
3. empty title
Screenshot 2026-08-12 at 17 49 28

and now you can see the same error in both cases, but as I understand we going to have another error for second case.

I have 2 options to solve this problem:
1: In this case we have diff error message for case if field is empty
Screenshot 2026-08-12 at 17 55 31
2. Or just follow description approach and have the same error for both cases
Screenshot 2026-08-12 at 17 58 23

what do you think about that?

dest411:
option 2 is definitely the best option
@MargoMatskevych 
good job

dest411 (approved):

